PR #7, opened 2025-09-02
feat: Add function calling to LLMs

d3xvn:


examples/07_function_calling_example/function_calling_example.py
```diff
@@ -0,0 +1,252 @@
...
+    await test_function_calling()
+    
+    print("\n" + "=" * 50)
+    print("🎥 Starting video call with function calling...")
+    
+    # Create Stream client
+    client = Stream.from_env()
+    
+    # Choose which agent to use
+    use_gemini = False  # Set to True to use Gemini Live instead
```
Nash0x7E2, line 181: Why not break this in to a different example?

examples/07_function_calling_example/function_calling_example.py
```diff
@@ -0,0 +1,252 @@
...
+
+class WeatherCondition(Enum):
+    """Enum for weather conditions."""
+    SUNNY = "sunny"
+    CLOUDY = "cloudy"
+    RAINY = "rainy"
+    SNOWY = "snowy"
+
+
+async def create_openai_agent_with_functions():
```
Nash0x7E2, line 32: Maybe made this a dedicated file just for OpenAI

examples/07_function_calling_example/function_calling_example.py
```diff
@@ -0,0 +1,252 @@
...
+    async def get_weather(location: str, unit: str = "celsius") -> str:
+        """
+        Get weather information for a specific location.
+        
+        Args:
+            location: The city or location to get weather for
+            unit: Temperature unit (celsius or fahrenheit)
+        """
+        # Simulate API call
+        await asyncio.sleep(0.1)  # Simulate network delay
```
Nash0x7E2, line 52: Don't love this, I think the examples should be simple and real-world so hitting a weather API would be more impactful than the aysnc timeout

examples/07_function_calling_example/function_calling_example.py
```diff
@@ -0,0 +1,252 @@
...
+        llm=llm,
+        tts=ElevenLabsTTS(),
+        stt=DeepgramSTT(),
+        turn_detection=FalTurnDetection(),
+    )
+    
+    return agent
+
+
+async def create_gemini_agent_with_functions():
```
Nash0x7E2, line 128: Consider moving to a dedicated file for Gemini. Keeping this in a single file overcomplicates the demo

examples/07_function_calling_example/function_calling_example.py
```diff
@@ -0,0 +1,252 @@
...
+    # Create the agent (STS mode - no separate STT/TTS needed)
+    agent = Agent(
+        edge=StreamEdge(),
+        llm=llm,
+    )
+    
+    return agent
+
+
+async def test_function_calling():
```
Nash0x7E2: Is this necessary?

examples/08_custom_agent_with_functions/pyproject.toml
```diff
@@ -0,0 +1,14 @@
...
+requires-python = ">=3.13"
+dependencies = [
+  "python-dotenv>=1.0",
+  "getstream[webrtc]==2.3.0a5",
+  "stream-agents",
+]
+
+[tool.uv.sources]
+"stream-agents" = {path = "../../stream-agents", editable=true}
+"getstream" = {path = "../../../stream-py", editable=true}
```
Nash0x7E2: Why are we making this a workspace path? For external users, this demo will break

examples/08_custom_agent_with_functions/simple_custom_agent.py
```diff
@@ -0,0 +1,122 @@
...
+
+from dotenv import load_dotenv
+from getstream import Stream
+from getstream.plugins import DeepgramSTT, ElevenLabsTTS
+from stream_agents.turn_detection import FalTurnDetection
+from stream_agents.llm import OpenAILLM
+from stream_agents import Agent, StreamEdge, start_dispatcher, open_demo
+
+
+class SimpleTaskAgent(Agent):
```
Nash0x7E2, line 18: Hmm, do we need this in the SDK itself or can this later be a guide/cookbook?

stream-agents/stream_agents/llm/README-function-calling.md
```diff
@@ -0,0 +1,167 @@
+# Function Calling System - RFC
+
+## Overview
```
Nash0x7E2: Nice that you're already writing docs for these :)

stream-agents/stream_agents/llm/openai.py
```diff
@@ -339,6 +343,11 @@ async def generate_chat(
         try:
             # Prepare parameters with defaults
             params = {"model": self._name, "messages": messages, **kwargs}
+            
+            # Add function tools if available and requested
+            if use_functions and self.function_registry._functions:
```
Nash0x7E2: What if I want to pass in tools manually?

stream-agents/stream_agents/turn_detection/__init__.py
```diff
@@ -5,6 +5,7 @@
     TurnDetection,
 )
 from .fal_turn_detection import FalTurnDetection
+from .krisp.krisp_turn_detection import KrispTurnDetection
```
Nash0x7E2: If you run into krisp errors, just remove this line from the export :)

examples/08_custom_agent_with_functions/simple_custom_agent.py
d3xvn: For now, I'd want it in the docs because it's two different ways to use it, but we can maybe have it inside the docs later.

stream-agents/stream_agents/llm/openai.py
d3xvn: Define manually?

d3xvn: If you mean params["tools"], you can't. Adds redundant functionality, and I don't think we should do it for now since we have a proper registry for it.

d3xvn: @maxkahan is this usually something you add in Python?

examples/07_function_calling_example/function_calling_example.py
d3xvn: It would be the same example, easier to test this way. Again, we can split this later, but I'd recommend letting it be for now.

examples/07_function_calling_example/function_calling_example.py
d3xvn: Easier to test without doing it all manually but fair point, removing this.

examples/08_custom_agent_with_functions/pyproject.toml
d3xvn: Yeah fair, we shouldn't have a workspace path for stream-py

stream-agents/stream_agents/llm/function_registry.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+class FunctionRegistry:
+    """Registry for LLM-callable functions with automatic schema inference."""
+    
+    def __init__(self):
+        self._functions: Dict[str, FunctionDefinition] = {}
+        self.logger = logging.getLogger("FunctionRegistry")
+    
+    def function(self, description: str = "", name: Optional[str] = None):
```
maxkahan: Might be good to align this naming with e.g. fastmcp, calling them tools etc

stream-agents/stream_agents/llm/function_registry.py
```diff
@@ -0,0 +1,323 @@
...
+            "type": "object",
+            "properties": properties
+        }
+        
+        if required:
+            schema["required"] = required
+            
+        return schema
+    
+    def _type_to_json_schema(self, type_hint: Any) -> Dict[str, Any]:
```
maxkahan: Feels like there should be a better way to handle this, this feels like a solved problem but I don't know the solution 🤷‍♂️

stream-agents/stream_agents/llm/function_registry.py
```diff
@@ -0,0 +1,323 @@
...
+                result = func_def.handler(**arguments)
+            
+            self.logger.debug(f"Function {name} executed successfully")
+            return result
+            
+        except Exception as e:
+            self.logger.error(f"Error executing function {name}: {e}")
+            raise
+    
+    def _validate_arguments(self, schema: Dict[str, Any], arguments: Dict[str, Any]) -> None:
```
maxkahan: This is the direct use case for Pydantic

stream-agents/stream_agents/llm/function_registry.py
```diff
@@ -0,0 +1,323 @@
+"""
```
maxkahan: This file is quite verbose

stream-agents/stream_agents/llm/function_registry.py
```diff
@@ -0,0 +1,323 @@
...
+        """Get list of registered function names."""
+        return list(self._functions.keys())
+    
+    def get_function_info(self, name: str) -> Optional[Dict[str, Any]]:
+        """Get information about a registered function."""
+        if name not in self._functions:
+            return None
+        
+        func = self._functions[name]
+        return {
```
maxkahan: Can't we return FunctionDefinition object we used to populate this?

stream-agents/stream_agents/llm/llm.py
```diff
@@ -1,59 +1,93 @@
...
-    while True:
-        bytestream = await audio_in_queue.get()
-        await audio_track.write(bytestream)
-
-'''
 
 
 class LLM:
-    sts: bool = False
+    """
```
maxkahan: You should sync with Thierry and make sure that you're using the updated code from llm2 branch

stream-agents/stream_agents/llm/llm.py
```diff
@@ -1,59 +1,93 @@
...
+    All LLM implementations should inherit from this class to get
+    automatic function calling capabilities.
+    """
+    
+    def __init__(self):
+        self.function_registry = FunctionRegistry()
+        self.sts: bool = False
+        self.logger = logging.getLogger(f"{self.__class__.__name__}")
+    
+    def function(self, description: str = "", name: Optional[str] = None):
```
maxkahan: Probably wouldn't advise calling anything "function", will get confusing

stream-agents/stream_agents/llm/llm.py
```diff
@@ -1,59 +1,93 @@
...
-
-    def attach_incoming_video(self, track):
-        pass
-
+        """Attach incoming audio track for processing."""
+        raise NotImplementedError("Subclasses must implement attach_incoming_audio")
+    
     def attach_outgoing_audio(self, track):
-        pass
+        """Attach outgoing audio track for responses."""
```
maxkahan: We usually talk about audio tracks, we should think about what we want to do with naming/presentation here

stream-agents/stream_agents/llm/openai.py
maxkahan: Need to decide how you want the user to interact with it. If the plan is to have a separate field at the top-level where the user defines this stuff, that's fine as long as we're able to expose the tools to the provider and get responses etc. with no additional user work.

maxkahan: Actually, having said that, if the user doesn't use the agents and can't define tools themselves to use that's a very bad thing, so this should be addressed. The user needs to be able to access the functionality without having to use Agent

stream-agents/stream_agents/llm/openai.py
```diff
@@ -384,6 +400,67 @@ async def generate_chat(
             self.logger.error(f"Error generating chat response: {e}")
             raise
 
+    async def _handle_function_calls(
```
maxkahan: Should have this in the base class as an abstract method

stream-agents/stream_agents/llm/openai.py
```diff
@@ -384,6 +400,67 @@ async def generate_chat(
...
+                )
+                
+                # Add function result to messages
+                messages.append({
+                    "role": "tool",
+                    "tool_call_id": call.id,
+                    "content": json.dumps(result)
+                })
+                
+            except Exception as e:
```
maxkahan: Should emit one of your error events 😄

stream-agents/stream_agents/llm/README-function-calling.md
```diff
@@ -0,0 +1,167 @@
...
+│   ├── OpenAI Tools Format
+│   └── Function Call Handling
+└── GeminiLiveModel
+    ├── Gemini Tools Format
+    └── Real-time Function Calls
+```
+
+## Implementation Details
+
+### Function Registry
```
maxkahan: We need a way to use without agent mode, please document

stream-agents/stream_agents/llm/README-function-calling.md
```diff
@@ -0,0 +1,167 @@
...
+```python
+from stream_agents import Agent
+
+class TaskAgent(Agent):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.tasks = []
+        self._register_functions()
+    
+    def _register_functions(self):
```
maxkahan: this feels strange

PR #8, opened 2025-09-02
Move plugins

maxkahan:


Nash0x7E2:
@maxkahan, When running the `simple_agent_example.py`, I get this error:

```shell 
error: Failed to generate package metadata for `krisp-audio==1.4.0 @ path+../../stream-agents/stream_agents/turn_detection/krisp/krisp_audio-1.4.0-cp313-cp313-linux_aarch64.whl`
  Caused by: Failed to read from the distribution cache
  Caused by: failed to query metadata of file `/Users/nash/git_projects/stream/video_ai/stream-agents/stream-agents/stream_agents/turn_detection/krisp/krisp_audio-1.4.0-cp313-cp313-linux_aarch64.whl`: No such file or directory (os error 2)
```

Nash0x7E2:
Pushed a fix for the Krisp issue. However running sync and the example `simple_agent_example.py` throws the following:

```shell 
(simple-agents-example) nash@Neevashs-MacBook-Pro 01_simple_agent_example % uv run simple_agent_example.py
Traceback (most recent call last):
  File "/Users/nash/git_projects/stream/video_ai/stream-agents/examples/01_simple_agent_example/simple_agent_example.py", line 7, in <module>
    from stream_agents.turn_detection import FalTurnDetection
  File "/Users/nash/git_projects/stream/video_ai/stream-agents/stream-agents/src/stream_agents/__init__.py", line 1, in <module>
    from .agents import Agent as Agent
  File "/Users/nash/git_projects/stream/video_ai/stream-agents/stream-agents/src/stream_agents/agents/__init__.py", line 7, in <module>
    from .agents import Agent as Agent
  File "/Users/nash/git_projects/stream/video_ai/stream-agents/stream-agents/src/stream_agents/agents/agents.py", line 9, in <module>
    from openai.types.responses import EasyInputMessageParam, ResponseInputItemParam
ModuleNotFoundError: No module named 'openai'
```

Looking at this next

examples/plugins/audio_moderation/main.py
```diff
@@ -0,0 +1,229 @@
...
+                if hasattr(event, "context") and event.context:
+                    print(f"    └─ context: {event.context}")
+
+            # Keep the connection alive and wait for audio
+            print("🎧 Listening for audio... (Press Ctrl+C to stop)")
+            await connection.wait()
+
+    except asyncio.CancelledError:
+        print("\n⏹️  Stopping moderation bot...")
+    except Exception as e:
```
yarikdevcom, line 181: not blocking but just in general big usage of except Exception should be covered with logger.exception so it automatically print traceback no need for traceback import

plugins/common/events.py
```diff
@@ -0,0 +1,560 @@
...
+        return str(self.error) if self.error else "Unknown error"
+
+
+# ============================================================================
+# Event Type Mappings for Easy Access
+# ============================================================================
+
+# Map event types to their corresponding classes
+EVENT_CLASS_MAP = {
+    EventType.STT_TRANSCRIPT: STTTranscriptEvent,
```
yarikdevcom, line 471: When we adding new plugin, we should update this class map? may-be some registry with decorator or function to add events from plugin itself? just thoughts

examples/plugins/audio_moderation/main.py
maxkahan: We use this pattern everywhere rn for speed but I would love to see an example of what you propose

plugins/common/events.py
maxkahan: Worth thinking about, again I'd be happy to see an example

Nash0x7E2:
LGTM. Once merged, let's move the base classes over to the Agents repo.

PR #13, opened 2025-09-04
Feat/vad events

Nash0x7E2:
Emit:
- Speech Start 
- Speech end 
- Inference

### DIFF
--- a/plugins/silero/vad/vad.py
+++ b/plugins/silero/vad/vad.py
@@ -1,5 +1,5 @@
 import logging
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 
 import torch
 import numpy as np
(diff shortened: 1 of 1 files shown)


plugins/common/events.py
```diff
@@ -361,6 +362,10 @@ class VADSpeechStartEvent(BaseEvent):
     speech_probability: float = 0.0
     activation_threshold: float = 0.0
     frame_count: int = 1
+    # Silero-specific enhancements
```
maxkahan: Might be better to subclass the events rather than adding silero-specific stuff to them, unless other vads have similar data you could expose the same way

plugins/silero/vad/vad.py
```diff
@@ -386,32 +435,194 @@ async def _flush_speech_buffer(self, user: Optional[Dict[str, Any]] = None) -> N
...
+
+            # Force flush if speech duration exceeds maximum
+            if self.total_speech_frames >= max_speech_frames:
+                await self._flush_speech_buffer(user)
 
     async def reset(self) -> None:
         """Reset the VAD state."""
         await super().reset()
         self.reset_states()
+        # Reset enhanced state tracking
```
maxkahan, line 646: Probably best to have methods to do resetting rather than repeating the code

PR #28, opened 2025-09-15
Bugfix/event current system

yarikdevcom:
- fixes current event system calls

## Summary by CodeRabbit

- New Features
  - Example app now initializes an explicit agent user for clearer, more personalized behavior.
- Behavior Changes
  - Removed the automatic response during call join in the example, giving you full control over call interactions.
- Refactor
  - Improved connection management reliability behind the scenes for more consistent runtime behavior.
- Documentation
  - Simplified example logs and flow for easier understanding and customization.
- Chores
  - Updated underlying dependency to the latest pinned revision for compatibility and stability.

### DIFF
--- a/examples/01_simple_agent_example/simple_agent_example.py
+++ b/examples/01_simple_agent_example/simple_agent_example.py
@@ -17,12 +17,11 @@ load_dotenv()
 async def start_agent() -> None:
     # create a stream client and a user object
     client = Stream.from_env()
-    #client.video.query_calls
-    #agent_user = client.create_user(name="My happy AI friend")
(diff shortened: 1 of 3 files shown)


examples/01_simple_agent_example/simple_agent_example.py
```diff
@@ -43,7 +43,7 @@ async def my_other_handler(participant):
         # TODO: wait till we have confirmation from client it can hear us
```
Nash0x7E2: Can you remove the commented out code if they are no longer required?

yarikdevcom: Make sense yes will do

PR #30, opened 2025-09-16
feat: new event system

yarikdevcom:
- managed by queue (so we can clean it manipulate)
- native async
- simple integration with current event data classes
- autogeneration of event statements to import
- simple registration
```

    @agent.subscribe
    async def my_handler(event: CallMemberAddedEvent):
        await agent.say(f"Hello, {event.type}")

```

- event registry that pick ups events automatically from `getstream.models`
- all events are typed
- you can add events with `append` instead of calling them directly
- recursive events supported, you can call send inside handler if needed

## Summary by CodeRabbit

- **New Features**
  - Per-instance event manager with subscribe/send APIs; Agent supports subscribe(), push-based event routing, say(text) using conversation+TTS, and send_audio(pcm).
  - Example: greeting-on-join handler added.

- **Breaking Changes**
  - join now requires a Call object.
  - Legacy global event registry and ReplyQueue removed; say no longer accepts user_id.

- **Chores**
  - Added getstream/getstream-plugin hints to example configs.

- **Tests**
  - Tests refocused on EventManager behavior and exception-event routing.

### DIFF
--- a/agents-core/pyproject.toml
+++ b/agents-core/pyproject.toml
@@ -12,8 +12,7 @@ dynamic = ["version"]
 requires-python = ">=3.10"
 dependencies = [
     # getstream[webrtc] currently comes from the latest commit on the webrtc branch in git
-    "getstream[webrtc] @ git+https://github.com/GetStream/stream-py.git@5d22ac964838da4f38af543dcbafd68f89be276a#egg=getstream",
-    # "getstream",
+    "getstream[webrtc] @ git+https://github.com/GetStream/stream-py.git@cebdd54#egg=getstream",
     "python-dotenv>=1.1.1",
     "pillow>=11.3.0",
     "aiortc>=1.13.0",


agents-core/stream_agents/core/events/manager.py
```diff
@@ -0,0 +1,102 @@
...
+        import_file.append("]")
+        import_file.append("")
+        return import_file
+
+    def subscribe(self, function):
+        subscribed = False
+        for name, event_class in function.__annotations__.items():
+            # check union of event classes (not neeeded for now)
+            if type(event_class) is types.UnionType or subscribed:
+                raise ValueError("Multiple events per handler don't supported")
```
maxkahan: "Multiple events per handler are not supported"

agents-core/stream_agents/core/agents/agents.py
```diff
@@ -197,6 +202,10 @@ def decorator(func):
             return func
         return decorator
 
+    def subscribe(self, function):
```
maxkahan, line 180: We should make what this does as clear as possible as it's a user-facing function

examples/01_simple_agent_example/simple_agent_example.py
```diff
@@ -30,15 +30,9 @@ async def start_agent() -> None:
...
-    @agent.on(EventType.PARTICIPANT_JOINED)
-    async def my_handler(participant):
-        await asyncio.sleep(5)
-        await agent.say(f"Hello, {participant.name}")
-        agent.logger.info(f"handled event {participant}")
-
-    @agent.on(EventType.CALL_MEMBER_ADDED)
-    async def my_other_handler(participant):
-        agent.logger.info(f"Call.* handled event {participant}")
+    @agent.subscribe
```
maxkahan, line 42: This is good, can you also add other examples of events from different places, e.g. the plugin events, in this example?

agents-core/stream_agents/core/events/__init__.py
```diff
@@ -51,7 +51,117 @@
 )
 from .events import create_event
 
+from getstream.models import (
```
maxkahan, line 27: I assume you're going to add similar visibility for the plugin events when they're done?

plugins/anthropic/stream_agents/plugins/anthropic/anthropic_llm.py
```diff
@@ -105,7 +105,10 @@ async def create_message(self, *args, **kwargs) -> LLMResponse:
             for msg in normalized_messages:
                 self._conversation.messages.append(msg)
 
-        self.emit("before_llm_response", self._normalize_message(kwargs["messages"]))
+        class BeforeLlmResponseEvent:
```
maxkahan, line 108: LLM not Llm

agents-core/stream_agents/core/events/__init__.py
yarikdevcom: Yes, each plugin now have events.py file with listed events, so you can subscribe per-plugin bases or core events basis

plugins/anthropic/stream_agents/plugins/anthropic/anthropic_llm.py
yarikdevcom: Yup, still fixing plugins, just finished with core one

PR #36, opened 2025-09-18
fix for example and startup/shutdown

yarikdevcom:
## Summary by CodeRabbit

- Refactor
  - CLI execution simplified: agents now run directly and the process exits when they complete. Background lifecycle and OS signal-based graceful shutdown handling removed. Improves predictability for short-lived runs; long-running sessions no longer wait on shutdown events.

- Chores
  - Updated the WebRTC-related dependency to a newer commit/revision.

### DIFF
--- a/agents-core/stream_agents/core/cli.py
+++ b/agents-core/stream_agents/core/cli.py
@@ -35,17 +35,6 @@ def setup_logging(log_level: str) -> None:
     logging.getLogger("test123").setLevel(numeric_level)
 
 
-def setup_signal_handlers() -> None:
-    """Set up signal handlers for graceful shutdown."""
-
-    def signal_handler(signum, frame):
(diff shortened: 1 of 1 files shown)


agents-core/stream_agents/core/cli.py
```diff
@@ -62,14 +62,15 @@ async def start_dispatcher(
     """
     # Setup logging and signal handlers
     setup_logging(log_level)
-    setup_signal_handlers()
+    #setup_signal_handlers()
```
maxkahan, line 65: Just remove?

yarikdevcom: yeah, doing, forgot to cleanup

yarikdevcom: done

PR #50, opened 2025-09-25
Fix instructions

maxkahan:
Fix instructions not being passed to openai from the agent mode
Add a jitter queue to help with audio issues with openai (still work to do here)

## Summary by CodeRabbit

- New Features
  - Enhanced realtime audio with buffered input for smoother playback and fewer dropouts.
  - On-the-fly resampling supports varied input sample rates.
  - Remote audio handling improved with dedicated background processing.
  - Sessions can include optional custom instructions for tailored behavior.

- Bug Fixes
  - Reduced audio glitches and timing issues; improved task cleanup to prevent leaks.

- Documentation
  - Updated example greeting text to clarify “next user input.”

- Refactor
  - Streamlined instruction handling during connection setup for more reliable initialization.


Nash0x7E2:
@maxkahan while testing your branch I noticed an interesting quirk with the Agent. If the call start with the camera on, the Agent is able to see the frames no problem. However, if the camera is off at the start and the user turn it on later in the call, the Agent is not able to see anything. 

For comparison, testing on `main` shows that the `Agent` is able to see the video correctly in both scenarios.

PR #55, opened 2025-09-26
Clean up plugin examples

Nash0x7E2:
The existing examples were using the `Stream-Py` lib directly. This PR updates them on how to use the Agent class. Still not sure if we need dedicated plugin examples. The value here is debatable. 

## Summary by CodeRabbit

- New Features
  - Introduced say() to trigger TTS responses.
  - Standardized streamed LLM events to LLMResponseChunk/Completed.
  - Auto-configure audio output from TTS (framerate/stereo).
  - Added Track Ended notifications and richer user metadata in media events.
  - open_demo is now async for more reliable demo launches.

- Refactor
  - Migrated multiple examples to an Agent-based, event-driven architecture (STT/TTS/VAD/moderation).

- Documentation
  - Updated guides to new event names; removed deprecated example docs.

- Tests
  - Updated to the new LLM streaming events.

- Chores
  - Fixed dependency typo; ignored .venv.


Nash0x7E2:
@yarikdevcom  I am blocked here on testing these changes because they depend on the event/stt changes.

yarikdevcom:
Got it, switching to this branch and digging into

agents-core/stream_agents/core/agents/agents.py
```diff
@@ -470,6 +469,23 @@ async def _on_agent_say(self, event: events.AgentSayEvent):
             )
             self.logger.error(f"Error in agent say: {e}")
 
+    async def say(self, text: str, user_id: Optional[str] = None, metadata: Optional[Dict[str, Any]] = None):
```
Nash0x7E2, line 472: Who do we need this? Should simple_response be used instead?

Nash0x7E2:
@yarikdevcom Before merging, lets:
1. Review the method addition to `Agent` I don't think it's required anymore 
2. Check that each example runs and works as expected lcoally 
3. Resolve the ruff/mypy issues

agents-core/stream_agents/core/agents/agents.py
yarikdevcom: say != simple_response, simple_response is input for LLM it's not response from agent, say means just say raw text

yarikdevcom: @maxkahan @Nash0x7E2 @tschellenbach  I propose to do it like agent.prompt("Do this and say that") agent.say("I'm agent")

PR #70, opened 2025-10-07
fix: Agent Example and TURN detection

d3xvn:
- Implement automatic LLM triggering in _on_transcript() for both modes:
  * Without turn detection: triggers immediately on transcript completion
  * With turn detection: accumulates transcripts and waits for TurnEndedEvent
- Add _pending_user_transcripts dict to track multi-chunk transcripts per user
- Implement turn detection LLM response in _on_turn_event()
- Add TTS interruption when user starts speaking (barge-in)
- Fix FAL turn detection event emission logic
- Fix double TTS triggering in OpenAI LLM plugin (was emitting LLMResponseCompletedEvent twice)

## Summary by CodeRabbit

* **New Features**
  * Turn-based handling with per-speaker transcript accumulation and optional immediate responses.
  * Automatic interruption of TTS when another participant begins speaking.
* **Bug Fixes**
  * Avoid duplicate "response completed" events for streaming LLM outputs.
* **Documentation**
  * Example updated to demonstrate enabling turn detection.
* **Chores**
  * Renamed dependencies from stream-agents to vision-agents plugins; added fal-client and updated package sources.

### DIFF
--- a/agents-core/vision_agents/core/agents/agents.py
+++ b/agents-core/vision_agents/core/agents/agents.py
@@ -794,7 +794,7 @@ class Agent:
             # Without turn detection: trigger LLM immediately on transcript completion
             # This is the traditional STT -> LLM flow
             if self.llm:
-                self.logger.info(f"🤖 Triggering LLM response immediately (no turn detection)")
+                self.logger.info("🤖 Triggering LLM response immediately (no turn detection)")
                 
                 # Get participant from event metadata
                 participant = None


agents-core/vision_agents/core/agents/agents.py
```diff
@@ -664,15 +667,51 @@ async def _process_track(self, track_id: str, track_type: str, participant):
     async def _on_turn_event(self, event: TurnStartedEvent | TurnEndedEvent) -> None:
         """Handle turn detection events."""
         if isinstance(event, TurnStartedEvent):
-            # TODO: Implement TTS pause/resume functionality
-            # For now, TTS will continue playing - this should be improved
-            self.logger.info(
-                f"👉 Turn started - participant speaking {event.speaker_id} : {event.confidence}"
-            )
+            # Interrupt TTS when user starts speaking (barge-in)
```
maxkahan, line 674: skip out of this earlier if we're in realtime mode

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -727,6 +766,33 @@ async def _on_transcript(self, event: STTTranscriptEvent | RealtimeTranscriptEve
...
+                self._pending_user_transcripts[user_id] += " " + event.text
+            
+            self.logger.debug(
+                f"📝 Accumulated transcript for {user_id} (waiting for turn end): "
+                f"{self._pending_user_transcripts[user_id][:100]}..."
+            )
+        else:
+            # Without turn detection: trigger LLM immediately on transcript completion
+            # This is the traditional STT -> LLM flow
+            if not self.realtime_mode and self.llm:
```
maxkahan: Same comment here about realtime

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -727,6 +766,33 @@ async def _on_transcript(self, event: STTTranscriptEvent | RealtimeTranscriptEve
...
+                self.logger.info(f"🤖 Triggering LLM response immediately (no turn detection)")
+                
+                # Get participant from event metadata
+                participant = None
+                if hasattr(event, "user_metadata"):
```
maxkahan, line 801: Slop?

d3xvn: I'm going to keep it as it is because it's a bit undefined but also because we can likely remove the whole participant later. It's not being used by simple_response at all.

PR #119, opened 2025-10-23
Fix screensharing

maxkahan:
1. Fixes bug with non-even numbers of pixels in screenshare causing it to fail
2. Fixes republished tracks not being shared with the AI
3. Tests for the above

## Summary by CodeRabbit

* **New Features**
  * Enhanced video track lifecycle management with support for track re-publishing and automatic fallback to alternative video sources when tracks are removed.
  * Upgraded video codec from VP8 to H.264.

* **Bug Fixes**
  * Improved video frame dimension handling for encoding compatibility.

* **Tests**
  * Added regression tests for track event flow and video dimension utilities.

* **Chores**
  * Enhanced logging verbosity in example configurations.

### DIFF
--- a/plugins/getstream/vision_agents/plugins/getstream/stream_edge_transport.py
+++ b/plugins/getstream/vision_agents/plugins/getstream/stream_edge_transport.py
@@ -22,7 +22,7 @@ from getstream.video.rtc.tracks import SubscriptionConfig, TrackSubscriptionConf
 
 from vision_agents.core.edge import EdgeTransport, sfu_events
 from vision_agents.plugins.getstream.stream_conversation import StreamConversation
-from vision_agents.core.edge.types import Connection, User
+from vision_agents.core.edge.types import Connection, User, OutputAudioTrack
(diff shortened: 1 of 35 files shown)


plugins/getstream/vision_agents/plugins/getstream/stream_edge_transport.py
```diff
@@ -270,13 +282,12 @@ async def on_track(track_id, track_type, user):
         self.events.silent(events.AudioReceivedEvent)
 
         @connection.on("audio")
-        async def on_audio_received(pcm: PcmData, participant: Participant):
+        async def on_audio_received(pcm: PcmData | None, participant: Participant):
```
d3xvn, line 285: Any reason we allow this to be none?

maxkahan: It's what the event expects, for the case where we want to log that audio is received but not actually do something with it directly. This just fixes the typing

tschellenbach (approved):

PR #154, opened 2025-11-05
Feat: Add support for Moondream VLM functions 

Nash0x7E2:
- Adds support for both Caption and VQA on Moondream 
- Add a basic example to the plugin dir 

## Summary by CodeRabbit

* **New Features**
  * Cloud and local vision–language models (VQA & captioning) with STT-triggered inputs, streaming responses, and a demo script to join calls.
  * Agent warmup now includes processor warmups.

* **Tests**
  * Integration tests added for cloud and local VQA and captioning flows.

* **Documentation**
  * Expanded Moondream docs and example README with quick-starts and usage guidance.

* **Chores**
  * Project manifest added; improved device selection (auto-detect/force_cpu) and import/path cleanups.

### DIFF
--- a/plugins/moondream/example/moondream_vlm_example.py
+++ b/plugins/moondream/example/moondream_vlm_example.py
@@ -4,7 +4,7 @@ from dotenv import load_dotenv
 
 from vision_agents.core import User, Agent, cli
 from vision_agents.core.agents import AgentLauncher
-from vision_agents.plugins import deepgram, getstream, vogent, elevenlabs, moondream
+from vision_agents.plugins import deepgram, getstream, elevenlabs, moondream
 from vision_agents.core.events import CallSessionParticipantJoinedEvent
 import os
 


agents-core/vision_agents/core/agents/agent_launcher.py
```diff
@@ -92,6 +92,13 @@ async def warmup(self, **kwargs) -> None:
             if agent.turn_detection and hasattr(agent.turn_detection, 'warmup'):
                 logger.debug("Warming up turn detection: %s", agent.turn_detection.__class__.__name__)
                 warmup_tasks.append(agent.turn_detection.warmup())
+
+            # Warmup processors
+            if agent.processors and hasattr(agent.processors, 'warmup'):
```
dangusev, line 97: `hasattr(agent.processors, 'warmup')` is probably never going to be true because it's a list
```suggestion
            if agent.processors:
```

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_cloud_vlm.py
```diff
@@ -0,0 +1,250 @@
...
+    Using the CloudVLM, you can send frames to the hosted Moondream model to perform either captioning or Visual queries.
+    The instructions are taken from the STT service and sent to the model along with the frame. Once the model has an output, the results are then vocalised with the supplied TTS service.
+
+    You can specify whether to use the caption endpoint or query (VQA).
+    """
+
+    def __init__(
+            self,
+            api_key: Optional[str] = None,
+            mode: str = "vqa",  # Default to VQA
```
dangusev: I suggest using `typing.Literal` here to get better autocomplete:

```python
from typing import Literal

...
mode: Literal["vqa", "caption"] = "vqa"
```

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_cloud_vlm.py
```diff
@@ -0,0 +1,250 @@
...
+            max_workers: int = 10,
+    ):
+        super().__init__()
+
+        self.api_key = api_key or os.getenv("MOONDREAM_API_KEY")
+        self.max_workers = max_workers
+        self.mode = mode
+        self._shutdown = False
```
dangusev: `self._shutdown` is not used anywhere

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_cloud_vlm.py
```diff
@@ -0,0 +1,250 @@
...
+            # Validate API key
+            if not self.api_key:
+                raise ValueError("api_key is required for Moondream Cloud API")
+
+            # Initialize cloud model
+            self.model = md.vl(api_key=self.api_key)
+            logger.info("✅ Moondream SDK initialized")
+
+        except Exception as e:
+            logger.exception(f"❌ Failed to load Moondream model: {e}")
```
dangusev, line 241:
```suggestion
            logger.exception(f"❌ Failed to load Moondream model")
```

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_local_vlm.py
```diff
@@ -0,0 +1,375 @@
...
+
+                self.events.send(LLMResponseChunkEvent(delta=caption))
+                self.events.send(LLMResponseCompletedEvent(text=caption))
+                logger.info(f"Moondream caption: {caption}")
+                self._is_processing = False
+                return LLMResponseEvent(original=caption, text=caption)
+            else:
+                logger.error(f"Unknown mode: {self.mode}")
+                self._is_processing = False
+                return None
```
dangusev: This should be validated at `__init__` and fail with ValueError instead of just logging it.

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_cloud_vlm.py
```diff
@@ -0,0 +1,250 @@
...
+                logger.info(f"Moondream VQA response: {answer}")
+                self._is_processing = False
+                return LLMResponseEvent(original=answer, text=answer)
+
+            elif self.mode == "caption":
+                # Moondream SDK returns {"caption": <generator>}, extract the generator
+                self._is_processing = True
+                result = self.model.caption(image, length="normal", stream=True)
+                stream = result["caption"]
+                caption = await asyncio.to_thread(self._consume_stream, stream)
```
dangusev: Use `self.executor` instead
Or even better - update `self._consume_stream` to always run inside `self.executor`

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_local_vlm.py
```diff
@@ -0,0 +1,375 @@
...
+            await self._prepare_moondream()
+
+    async def _prepare_moondream(self):
+        """Load the Moondream model from Hugging Face."""
+        logger.info(f"Loading Moondream model: {self.model_name}")
+        logger.info(f"Device: {self.device}")
+        
+        # Load model in thread pool to avoid blocking event loop
+        # Transformers handles downloading and caching automatically via Hugging Face Hub
+        self.model = await asyncio.to_thread(  # type: ignore[func-returns-value]
```
dangusev, line 106: use `self.executor` or make `self._load_model_sync()` run inside executor instead (better option)

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_local_vlm.py
```diff
@@ -0,0 +1,375 @@
...
+            frame_array = latest_frame.to_ndarray(format="rgb24")
+            image = Image.fromarray(frame_array)
+
+            if self.mode == "vqa":
+                if not text:
+                    logger.warning("VQA mode requires text/question")
+                    return None
+
+                self._is_processing = True
+                result = await asyncio.to_thread(self.model.query, image, text, stream=True)
```
dangusev, line 265: use `self.executor` or add a method `self._model_query` that always does it in the thread

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_local_vlm.py
```diff
@@ -0,0 +1,375 @@
...
+
+                if isinstance(result, dict) and "answer" in result:
+                    stream = result["answer"]
+                else:
+                    stream = result
+
+                answer = await asyncio.to_thread(self._consume_stream, stream)
```
dangusev, line 272: use `self.executor`

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_cloud_vlm.py
```diff
@@ -0,0 +1,250 @@
...
+        result = "".join(chunks)
+        logger.debug(f"Moondream stream result: {result}")
+        return result
+
+    async def _process_frame(self, text: Optional[str] = None) -> Optional[LLMResponseEvent]:
+        if self._latest_frame is None:
+            logger.warning("No frames available, skipping Moondream processing")
+            return None
+
+        if self._is_processing:
```
dangusev: What's the end goal with `self._is_processing` here?
Can we use the `asyncio.Lock()` around the relevant blocks of code instead?

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_local_vlm.py
```diff
@@ -0,0 +1,375 @@
...
+        """Stop video forwarding."""
+        if self._video_forwarder is not None:
+            await self._video_forwarder.stop()
+            self._video_forwarder = None
+            logger.info("Stopped video forwarding")
+
+    def close(self):
+        """Clean up resources."""
+        self._shutdown = True
+        if hasattr(self, "executor"):
```
dangusev:
```suggestion
        if self.executor is not None:
```

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_cloud_vlm.py
Nash0x7E2: Ahh did not know you can do this. I had a note to create an enum to solve this but literal seems a lot cleaner, thanks!

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_cloud_vlm.py
Nash0x7E2: The goal here was the model would start responding and we'd continue to send frames before the full response was received. This created a situation where the model would send 2 - 3 responses and flood the TTS service. The is processing step was meant to ignore incoming text if we're already getting a response from the model.

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_cloud_vlm.py
Nash0x7E2: to_thread calls the executor internally

agents-core/vision_agents/core/agents/agent_launcher.py
Nash0x7E2: Not all processors will have the warmup method so we need this check here. For example, YOLO does not since the model is handled internally by YOLO but Moondream does because we're required to handle the model loading externally.

plugins/moondream/vision_agents/plugins/moondream/vlm/moondream_local_vlm.py
Nash0x7E2: to_thread does this under the hood

PR #100, opened 2025-10-13
Apply & update ruff in pre-commit

ulgens:
I found the repo thanks to [José Ramalho](https://www.linkedin.com/in/josencramalho/) reaching out to me over LinkedIn. I was looking to inspect the repo and test the tool, but I realized pre-commit hooks are outdated and they are not enforced in CI. 

I applied only ruff fixes and updated the pre-commit hook, but still it created a huge diff. If this gets merged, I can create 2 more PRs for `pre-commit-hooks` and `local`, then add a CI workflow to enforce the pre-commit checks - similar to [#1](https://github.com/ulgens/django-blasphemy/blob/574327dc1ae6742552a17b0b97384864b0c81a05/.github/workflows/prek.yml) or [#2](https://github.com/ulgens/djangoproject.com/blob/11d339ae951910035b2c152996217852bc42843b/.github/workflows/prek.yml)

---

Because this PR only focuses on ruff fixes and update, and coderabbit was generating a misleading summary, I deleted the automated AI summary from PR description.

PR #197, opened 2025-11-17
CI improvements

dangusev:
- Separate ruff and mypy runs for visibility
- use `--frozen` flag for uv to keep uv.lock intact. It should also improve UV cache hits, speeding up the CI
- update ruff version in precommit hooks

## Summary by CodeRabbit

* **Chores**
  * Enabled frozen install mode, set default Python to 3.13.9, added prune-cache and cache-suffix options, removed prior venv cache step, set default working directory, and clarified the dependency-install step name.

* **CI**
  * Added a UV_FROZEN flag, split lint into separate Ruff and mypy jobs, introduced mypy caching, separated core and plugin test runs, adjusted step names, and removed environment-printing steps.

* **Style**
  * Bumped Ruff hook, updated local mypy hook commands to use frozen mode and revised paths, and switched dev tooling to mypy[mypyc,faster-cache].

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -40,17 +40,17 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: .mypy_cache
-          key: mypy-cache-${{ hashFiles('pyproject.toml', 'mypy.ini') }}
+          key: mypy-cache-${{ hashFiles('pyproject.toml') }}
       - name: Run mypy for core
         run: uv run mypy --install-types --non-interactive agents-core/vision_agents
       - name: Run mypy for plugins
         run: uv run mypy --install-types --non-interactive --exclude 'plugins/.*/tests/.*' plugins
-      # Save the updated cache
+      # Save the updated mypy cache
       - name: Save mypy cache
         uses: actions/cache/save@v4
         with:
           path: .mypy_cache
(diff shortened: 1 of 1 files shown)


ulgens:
> Separate ruff and mypy runs for visibility

I'd recommend running pre-commit/prek for handling linters instead of giving each one a new wrapper. If the individual visibilities of the linters are important, they can still be called separately but via pre-commit/prek

Also, this seems related to https://github.com/GetStream/Vision-Agents/pull/100 . Would be nice to have a review together.

dangusev:
Hey @ulgens, thanks for keeping an eye on the PRs 👍 .

In the description, I meant separating Ruff and Mypy runs only in CI/CD. 
So if one of these checks fails, it's easier to spot which one in the logs.

The precommit hooks work the same as before.

ulgens:
@dangusev I mean

```
# Current version
uv run mypy --install-types --non-interactive --exclude 'plugins/.*/tests/.*' plugins
# vs
# Recommended version
prek (or pre-commit) run mypy
```

and 

```
# Current version
uv run ruff check .
vs
# Recommended version
prek (or pre-commit) run ruff
```

This approach allows to use the same config in multiple places without redefining parameters and dependencies/versions. It decreases the fragmentation, makes the CI config easier to understand and maintain.

PR #235, opened 2025-12-04
fix: subscribe to existing participant tracks when agent joins late

d3xvn:
When the agent joins a call after users are already present, the TrackPublished events were already sent before the agent connected. This caused the agent to miss video tracks from existing participants.

The fix manually triggers subscriptions for existing participants' published tracks after the connection is established, then emits TrackAddedEvent once the WebRTC tracks arrive.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed issue where the agent could not subscribe to media tracks from participants who joined before it connected.
  * Improved track synchronization and event handling so audio/video streams from existing participants are discovered, matched, and processed more reliably, with safer timeout and error handling when delayed or missing tracks.

### DIFF
--- a/plugins/getstream/vision_agents/plugins/getstream/stream_edge_transport.py
+++ b/plugins/getstream/vision_agents/plugins/getstream/stream_edge_transport.py
@@ -136,24 +136,25 @@ class StreamEdge(EdgeTransport):
                 except Exception as e:
                     logger.error(f"Failed to subscribe to existing track: {e}")
 
-        # Wait for WebRTC tracks to arrive after subscription
-        if tracks_to_subscribe:
-            await asyncio.sleep(1.0)
+        # Poll for WebRTC tracks to arrive after subscription
+        for participant, track_type_int in tracks_to_subscribe:
+            expected_kind = self._get_webrtc_kind(track_type_int)
+            track_key = (
+                participant.user_id,
+                participant.session_id,
+                track_type_int,
+            )
 
-            # Emit TrackAddedEvent for each track that arrived
-            for participant, track_type_int in tracks_to_subscribe:
-                expected_kind = self._get_webrtc_kind(track_type_int)
-                track_key = (
-                    participant.user_id,
-                    participant.session_id,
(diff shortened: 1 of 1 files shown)


plugins/getstream/vision_agents/plugins/getstream/stream_edge_transport.py
```diff
@@ -316,6 +399,9 @@ async def on_audio_received(pcm: PcmData):
         )  # TODO: weird API? there should be a manual version
         self._connection = connection
 
+        # Subscribe to tracks from participants who joined before the agent
+        await self._subscribe_to_existing_tracks(connection)
```
maxkahan, line 410: is this too coupled to the workflow we have atm? i.e. one user, one ai

plugins/getstream/vision_agents/plugins/getstream/stream_edge_transport.py
```diff
@@ -105,6 +105,89 @@ def _get_webrtc_kind(self, track_type_int: int) -> str:
...
+
+                try:
+                    await subscription_manager.handle_track_published(mock_event)
+                    tracks_to_subscribe.append((participant, track_type_int))
+                except Exception as e:
+                    logger.error(f"Failed to subscribe to existing track: {e}")
+
+        # Wait for WebRTC tracks to arrive after subscription
+        if tracks_to_subscribe:
+            await asyncio.sleep(1.0)
```
maxkahan: is there a way of doing this where we don't rely on a sleep?

d3xvn: Uses polling now instead.

plugins/getstream/vision_agents/plugins/getstream/stream_edge_transport.py
d3xvn: It will work for multiple participants as well. It may break for multiple agents but I feel like we haven't covered that use case in most of the SDK atm.

PR #237, opened 2025-12-05
Add Deepgram TTS capabilities

DaemonLoki:
## Release Notes

* **New Features**
  * Added Text-to-Speech (TTS) support to Deepgram plugin with streaming audio generation
  * Added example demonstrating TTS integration with other services

* **Documentation**
  * Updated plugin documentation with separate TTS and STT guidance
  * Added setup instructions and usage examples
  * Included available voices reference

* **Tests**
  * Added TTS integration tests

### DIFF
--- a/plugins/deepgram/vision_agents/plugins/deepgram/tts.py
+++ b/plugins/deepgram/vision_agents/plugins/deepgram/tts.py
@@ -1,4 +1,3 @@
-import asyncio
 import logging
 import os
 from typing import AsyncIterator, Optional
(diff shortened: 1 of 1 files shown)


plugins/deepgram/example/README.md
```diff
@@ -0,0 +1,48 @@
...
+2. Install dependencies:
+
+```bash
+uv sync
+```
+
+3. Run the example:
+
+```bash
+uv run python deepgram_tts_example.py dev --call-type audio_room --call-id test
```
Nash0x7E2, line 29: @DaemonLoki you don't need `uv run python` here, `uv run ...` is fine :)

PR #250, opened 2025-12-10
fix: sync partial transcripts to chat when turn ends

d3xvn:
Previously, chat messages were lost when only STTPartialTranscriptEvent events were received and no final STTTranscriptEvent arrived before the turn ended. The transcript was sent to the LLM but never synced to chat.

This fix moves chat sync from the STTTranscriptEvent handler to _on_turn_event, ensuring the accumulated transcript (from partials or finals) is always recorded in chat when a turn completes.

## Refactor
* Improved transcript synchronization to ensure consistent chat history updates for both final and partial transcripts.

### DIFF
--- a/agents-core/vision_agents/core/agents/agents.py
+++ b/agents-core/vision_agents/core/agents/agents.py
@@ -339,25 +339,9 @@ class Agent:
         async def on_error(event: STTErrorEvent):
             self.logger.error("stt error event %s", event)
 
-        @self.events.subscribe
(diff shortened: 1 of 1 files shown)


Nash0x7E2:
Hi @d3xvn, running the simple agent example on this branch produced the following:

| Example 1      | Example 2 |
| ----------- | ----------- |
|  CleanShot 2025-12-10 at 11 34 38@2x    |    CleanShot 2025-12-10 at 11 34 27@2x    |

PR #262, opened 2025-12-15
Processors clean up

dangusev:
- Removed unused code
- Converted `Processor` to a proper `ABC`. Added other base classes with different capabilities
- Fixed typing
- Fixed `YOLOPoseProcessor` errors on exit

## Summary by CodeRabbit

* **New Features**
  * Clearer processor roles: distinct audio/video processor and publisher types.
  * Agent sessions support improved for richer processor/session integration.
  * Video-only mode explicitly supported without requiring an LLM.

* **Bug Fixes**
  * Tighter configuration validation for audio/video/LLM combinations.

* **Refactor**
  * Major overhaul of processor API to simpler, type-safe interfaces; image-processing paths removed.
  * Processors now attach via a unified lifecycle method.

* **Tests**
  * Tests updated to the new public processor interfaces.

### DIFF
--- a/agents-core/vision_agents/core/agents/agents.py
+++ b/agents-core/vision_agents/core/agents/agents.py
@@ -168,7 +168,7 @@ class Agent:
         self.stt = stt
         self.tts = tts
         self.turn_detection = turn_detection
-        self.processors = processors or []
+        self.processors: list[Processor] = processors or []
         self.mcp_servers = mcp_servers or []
         self._call_context_token: CallContextToken | None = None
         self._context_token: Token[Context] | None = None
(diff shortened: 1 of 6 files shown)


Nash0x7E2 (commented):
@dangusev noticed you removed the Mixins for Audio processor and video processor. Would the idea be to just sub-class the Audio and Video publishers instead to get access to and transform the various tracks?

dangusev:
> @dangusev noticed you removed the Mixins for Audio processor and video processor. Would the idea be to just sub-class the Audio and Video publishers instead to get access to and transform the various tracks?

Yes, mixins became standalone classes that can be used as-is, and they also force specific methods to be implemented.
They can be subclassed similarly to mixins too

PR #271, opened 2025-12-24
fix screensharing when screen height/width is an odd pixel number

maxkahan:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced even video frame dimensions to ensure reliable H.264 encoding and playback.

* **Improvements**
  * Added clearer logging when video or screen-share tracks are added or removed for better observability.
  * Optimized cropping to remove a single pixel when needed, preserving performance and avoiding rescaling.

* **Tests**
  * Added unit tests validating even-dimension handling and correct cropping behavior.

* **Chores**
  * Minor import and parameter-name cleanups with no user-facing behavior changes.

### DIFF
--- a/tests/test_video_utils.py
+++ b/tests/test_video_utils.py
@@ -32,24 +32,6 @@ class TestEnsureEvenDimensions:
         assert result.height == 100
         assert result is frame  # Should be same object
 
-    def test_odd_width_cropped(self):
-        """Frame with odd width should be cropped by 1 pixel."""
-        frame = self._create_frame(101, 100)
-        result = ensure_even_dimensions(frame)
-
-        assert result.width == 100
-        assert result.height == 100
-        assert result is not frame
-
-    def test_odd_height_cropped(self):
-        """Frame with odd height should be cropped by 1 pixel."""
-        frame = self._create_frame(100, 101)
-        result = ensure_even_dimensions(frame)
-
-        assert result.width == 100
-        assert result.height == 100
-        assert result is not frame
-
     def test_both_odd_cropped(self):
         """Frame with both odd dimensions should be cropped."""
         frame = self._create_frame(101, 101)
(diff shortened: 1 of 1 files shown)


agents-core/vision_agents/core/utils/video_track.py
```diff
@@ -209,6 +205,9 @@ async def recv(self) -> VideoFrame:
         loop = asyncio.get_running_loop()
         frame = await loop.run_in_executor(self._executor, self._next_frame)
 
+        # Ensure even dimensions for H.264 compatibility
+        frame = ensure_even_dimensions(frame)
+
```
dangusev: Any particular reason to call `ensure_even_dimensions` on `recv()`? 
If there's none, we should not do that.

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -971,6 +971,15 @@ async def _on_track_change(self, track_id: str):
...
 
+        source_type = (
+            "SCREEN_SHARE"
+            if source_track.type == TrackType.TRACK_TYPE_SCREEN_SHARE
+            else "VIDEO"
+        )
+        logger.info(
+            f"📺 Source track selected: {source_type} (priority={source_track.priority})"
+        )
+
```
dangusev: I don't think we need these logs here because `_on_track_added` already prints them and then calls `_on_track_change`.

Adding them to `_on_track_removed` would make more sense in this case.

Nash0x7E2:
@maxkahan what's the status of this PR?

agents-core/vision_agents/core/utils/video_track.py
maxkahan: removed

maxkahan:
@dangusev @Nash0x7E2 made Daniil's edits, ready for re-review

agents-core/vision_agents/core/agents/agents.py
maxkahan: done

PR #273, opened 2025-12-29
New warmup implementation

dangusev:
## What's changed
- Added `Warmable` and `WarmupCache` to load resources once and share them between instances of the same classes in runtime.
- Updated `SileroVAD` wrapper to separate the model (which can be shared between calls) and the track state (which must be isolated).
- Updated smart_turn, vogent and `aws.Realtime`
- Enabled Vogent tests

## New `warmup()` approach.
Previously, some plugins had `warmup()` methods defined, which were called by `AgentLauncher`.
However, there was no way to share the warmed-up state between different instances of the same class.
E.g., a warmed-up TurnDetection would load the model every time a new call starts.

We need the warmup to be done only once during the lifetime of the program.

This PR adds `Warmable` and `WarmupCache` to standardize how the warmup is done and shared within the same process.

How it works:

1. `AgentLauncher` holds a single instance of `WarmupCache`, which stores the warmup data like VAD models, etc.
2.  `Warmable` objects define two methods:
  - `on_warmup` - called only once for the same `WarmupCache` instance for the same **class**. Here you load the resources and return them
  - `on_warmed_up` - called every time a new Agent is spawned. Here you receive the warmed up resources and set them back to your object.
3. `AgentLauncher.launch()` handles the warmup process under the hood.

**Example:** 

```python
from vision_agents.core.warmup import Warmable, WarmupCache

# Subclass `Warmable` ABC first and implement necessary methods
class SomeWarmable(Warmable[dict]):
    def __init__(self):
        self._resource: dict | None = None

    async def on_warmup(self):
        # This method is called once 
        # Do some loading work here and return some resource (eg. a dict with data)
        resource = {}
        return resource

    def on_warmed_up(self, resource: dict) -> None:
        # Set the resource back to the instance.
        self._resource = resource

````

## Summary by CodeRabbit

* **New Features**
  * Central warmup framework for coordinated, idempotent resource loading.
  * Shared VAD session pool to improve VAD efficiency and pooling.
  * Helper to run sync or async callables transparently.

* **Refactor**
  * Components adopt a unified warmup lifecycle (on_warmup/on_warmed_up); agents warm dependencies before launch.
  * Agent launcher now ensures warmup is performed prior to agent creation.

* **Tests / Docs**
  * Added warmup unit tests, updated integration tests, and documentation on plugin warmup.

### DIFF
--- a/plugins/smart_turn/vision_agents/plugins/smart_turn/smart_turn_detection.py
+++ b/plugins/smart_turn/vision_agents/plugins/smart_turn/smart_turn_detection.py
@@ -149,7 +149,7 @@ class SmartTurnDetection(
         ],
     ) -> None:
         vad_pool, whisper_extractor, smart_turn = resource
-        self._vad_session = vad_pool.session()
+        self._vad_session = vad_pool.session(self.vad_reset_interval_seconds)
         self._whisper_extractor = whisper_extractor
         self._smart_turn = smart_turn
 
--- a/plugins/vogent/vision_agents/plugins/vogent/vogent_turn_detection.py
+++ b/plugins/vogent/vision_agents/plugins/vogent/vogent_turn_detection.py
@@ -152,7 +152,7 @@ class VogentTurnDetection(
         self, resources: tuple[SileroVADSessionPool, WhisperModel, VogentDetector]
     ) -> None:
         vad_pool, self._whisper, self._vogent = resources
-        self._vad_session = vad_pool.session()
+        self._vad_session = vad_pool.session(self.vad_reset_interval_seconds)
 
     async def start(self):
         """

### CODE REVIEW

File: agents-core/vision_agents/core/utils/utils.py [hunk 1]
```diff
@@ -81,3 +86,16 @@ def write_file():
             raise RuntimeError(f"Failed to download {model_name}: {e}")
 
     return path
+
+
+async def await_or_run(
```
Nash0x7E2, line 91: this is nice :)

dangusev:
> @dangusev looks good, please double check our internal docs and add something on how the warmup/warmable flow works since it's a departure from the current norm everyone would be familiar with. Other than that, LGTM

Good point @Nash0x7E2 .
I added a paragraph about warmup to `DEVELOPMENT.md`.

PR #279, opened 2026-01-06
Close idle agents after some timeout & Agent clean up

dangusev:
### Main changes
1.  `AgentLauncher` now keeps track of the active agents and closes the idle ones after a specified `agent_idle_timeout` (default 30s).
  - The agent is considered **idle** if it's on the call and there's nobody else except the agent.
  - The idle timeout starts only after the `Agent.join()` is complete (the agent joins the RTC connection earlier).

2. `Agent.join()` now operates as an async context manager for improved resource management and cleanup
```python
# Used to be like this
with await agent.join(...):
    ...

# Now it's like this
async with agent.join(...):
...
```

All examples have been updated to use the new API.

### Other changes
  * `Agent.wait_for_participant()` now accepts a timeout (default 10s) to limit the waiting time.  
  When timeout expires, it will proceed and join the call.
    * The `wait_for_participant()` logic is moved to `StreamConnection` since it's tightly coupled with `getstream` implementation.
  * Fixed `Agent.close()` getting cancelled in the middle of a cleanup.
  * Fixed `elevenlabs.STT` and `deepgram.STT` cancelled in the middle of a cleanup, too.
  * StreamConnection and StreamEdge now track idle time and support waiting for participants
  * Updated all examples to use new async context manager syntax for `Agent.join()`
  * Fixed issues with `twilio` and `turbopuffer` introduced in 0.2.7 along the way 

## Summary by CodeRabbit

* **New Features**
  * Agent.join now uses async context-manager semantics; added wait_for_participant(), idle_for(), and closed property; cancel-and-wait helper added; launcher auto-cleans idle agents.

* **Bug Fixes**
  * More robust shutdown, task cancellation, connection close handling, and serialized join/close flows.

* **Documentation**
  * Examples and READMEs updated to use the async join pattern.

* **Tests**
  * Added AgentLauncher tests; removed an old agent test suite.

* **Chores**
  * Added Twilio and Turbopuffer workspace plugins.

### DIFF
--- a/agents-core/vision_agents/core/agents/agent_launcher.py
+++ b/agents-core/vision_agents/core/agents/agent_launcher.py
@@ -24,7 +24,7 @@ class AgentLauncher:
         self,
         create_agent: Callable[..., "Agent" | Awaitable["Agent"]],
         join_call: Callable[..., None | Awaitable[None]] | None = None,
-        agent_idle_timeout: float = 30.0,
+        agent_idle_timeout: float = 60.0,
         agent_idle_cleanup_interval: float = 5.0,
     ):
         """
@@ -33,7 +33,7 @@ class AgentLauncher:
         Args:
             create_agent: A function that creates and returns an Agent instance
             join_call: Optional function that handles joining a call with the agent
-            agent_idle_timeout: Optional timeout in seconds for agent to stay alone on the call. Default - `30.0`.
+            agent_idle_timeout: Optional timeout in seconds for agent to stay alone on the call. Default - `60.0`.
                 `0` means idle agents won't leave the call until it's ended.
 
         """
--- a/uv.lock
+++ b/uv.lock
@@ -6527,6 +6527,12 @@ roboflow = [
 smart-turn = [
     { name = "vision-agents-plugins-smart-turn" },
 ]
+turbopuffer = [
+    { name = "vision-agents-plugins-turbopuffer" },
+]
+twilio = [
+    { name = "vision-agents-plugins-twilio" },
+]
 ultralytics = [
     { name = "vision-agents-plugins-ultralytics" },
 ]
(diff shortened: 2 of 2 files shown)


agents-core/vision_agents/core/agents/agent_launcher.py
```diff
@@ -23,33 +24,72 @@ def __init__(
         self,
         create_agent: Callable[..., "Agent" | Awaitable["Agent"]],
         join_call: Callable[..., None | Awaitable[None]] | None = None,
+        agent_idle_timeout: float = 30.0,
```
Nash0x7E2: Should this be higher, maybe a minute?

examples/04_football_commentator_example/football_commentator_example.py
```diff
@@ -64,7 +64,7 @@ async def join_call(agent: Agent, call_type: str, call_id: str, **kwargs) -> Non
     call = await agent.create_call(call_type, call_id)
 
     # Have the agent join the call/room
-    with await agent.join(call):
+    async with agent.join(call):
```
Nash0x7E2, line 67: Dumb question, what's the difference between the two syntax?

Nash0x7E2:
Also remember to update visionagents.ai

examples/04_football_commentator_example/football_commentator_example.py
dangusev: `with await` means that the context manager object returned by `await` is synchronous.
Its `__enter__` and `__exit__` methods are not async.
And it's also not common overall to have it this way.

`async with` means that the context manager is async, its enter-exit methods are async too, and we can easily do `await` inside them.

In this particular case, this change allowed me to drop the whole `agent_session.py` file.

agents-core/vision_agents/core/agents/agent_launcher.py
dangusev: Sure, I don't see why not.

PR #280, opened 2026-01-07
Add support for Nemotron ASR

Nash0x7E2:
- NeMo has dependency conflicts with getstream, vogent, and protobuff 
- Solution so far is to run NeMo/model loading as a separate service/fast api 
- Then call that local server from VA for transcriptions 

cc: @dangusev


Nash0x7E2:
@dangusev can use a sanity check on this one when you have a moment, not sure if there's a better way to go about this but seems like NeMo (the library to load the model) has some tight constraints conflicting with multiple plugins

Nash0x7E2:
@dangusev let me know what you think here, maybe we just close for now

PR #325, opened 2026-01-29
feat: add custom events and metrics broadcasting via Stream Video

d3xvn:
Add ability to send custom events and metrics to call participants through Stream Video's custom events API. This enables building real-time dashboards in client apps.

- Add send_custom_event() to EdgeTransport and StreamEdge
- Add send_custom_event() and send_metrics_event() to Agent
- Add broadcast_metrics and broadcast_metrics_interval constructor params for automatic periodic metrics broadcasting
- Store Call reference in StreamEdge for event sending

## Release Notes

* **New Features**
  * Agents can now broadcast performance metrics to call participants with configurable intervals.
  * Added public APIs for sending custom events and metric snapshots to all connected participants.
  * New configuration options to enable metrics broadcasting and control broadcast frequency.

* **Tests**
  * Added comprehensive test coverage for metrics broadcasting and custom event functionality.

### DIFF
--- a/agents-core/vision_agents/core/agents/agents.py
+++ b/agents-core/vision_agents/core/agents/agents.py
@@ -143,6 +143,9 @@ class Agent:
         options: Optional[AgentOptions] = None,
         tracer: Tracer = trace.get_tracer("agents"),
         profiler: Optional[Profiler] = None,
+        # Metrics broadcasting to call participants
+        broadcast_metrics: bool = False,
+        broadcast_metrics_interval: float = 5.0,
     ):
         self._agent_user_initialized = False
         self.agent_user = agent_user
@@ -236,6 +239,11 @@ class Agent:
         self._video_track: Optional[VideoStreamTrack] = None
 
         self._audio_consumer_task: Optional[asyncio.Task] = None
+        self._metrics_broadcast_task: Optional[asyncio.Task] = None
+
+        # Metrics broadcasting settings
+        self._broadcast_metrics = broadcast_metrics
+        self._broadcast_metrics_interval = broadcast_metrics_interval
 
         # validation time
         self._validate_configuration()
@@ -589,6 +597,13 @@ class Agent:
             self._audio_consumer_task = asyncio.create_task(
                 self._consume_incoming_audio()
             )
+
+            # Start metrics broadcast if enabled
+            if self._broadcast_metrics:
+                self._metrics_broadcast_task = asyncio.create_task(
+                    self._metrics_broadcast_loop()
+                )
+
             self._call_ended_event = asyncio.Event()
             self._joined_at = time.time()
             yield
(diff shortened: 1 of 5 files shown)


agents-core/vision_agents/core/agents/agents.py
```diff
@@ -143,6 +143,9 @@ def __init__(
         options: Optional[AgentOptions] = None,
         tracer: Tracer = trace.get_tracer("agents"),
         profiler: Optional[Profiler] = None,
+        # Metrics broadcasting to call participants
+        broadcast_metrics: bool = False,
```
Nash0x7E2, line 147: Do we need this to be a public method? Why not a CLI option which would be cleaner? (Worried about bloating the public API)

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -143,6 +143,9 @@ def __init__(
...
+        broadcast_metrics_interval: float = 5.0,
```
Nash0x7E2, line 148: Same here, would a CLI option be a bit cleaner?

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -1417,6 +1456,60 @@ async def _get_video_track_override(self) -> VideoFileTrack:
     def metrics(self) -> AgentMetrics:
         return self._metrics
 
+    async def send_custom_event(self, data: dict) -> None:
```
Nash0x7E2, line 1459: Do we want this to be a public method? I understand why it is atm however as we're implementing support for local devices on the "local" branch, we may be leaking Stream-specific concepts (call.on for instance) to the Agent. Maybe give this an extra consideration.

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -1417,6 +1456,60 @@ async def _get_video_track_override(self) -> VideoFileTrack:
...
+        Example:
+            await agent.send_custom_event({
+                "type": "status_update",
+                "status": "processing",
+                "progress": 0.5
+            })
+        """
+        await self.edge.send_custom_event(data)
+
+    async def send_metrics_event(
```
Nash0x7E2, line 1481: Why the need to be public?

agents-core/vision_agents/core/agents/agents.py
d3xvn: Yes, this can be used by people for broadcasting their own metrics so it's definitely not Stream specific

agents-core/vision_agents/core/agents/agents.py
d3xvn: We can make this choice either way but I think it's easier if we also allow people to send their custom metrics if needed. There could be an argument made to change send_custom_event to private but we can change this later.

agents-core/vision_agents/core/agents/agents.py
d3xvn: The implementation is through the edge, so we're not leaking any other concepts or dependencies unless it's in the edge itself. We could make this public or private. For now I think it's easier if people want to send their events or metrics themselves, but again, we can change this if needed.

agents-core/vision_agents/core/agents/agents.py
d3xvn: I don't think a CLI option is a good one. I did consider having a method instead for example `start_broadcast()` but in the end I settled on this because it allows discoverability of this config option as well.

PR #330, opened 2026-02-03
Decouple vision agents from getstream

dangusev:
**Goal** - reduce dependency on `getstream` library across vision_agents (except `plugins/getstream`)

## Summary by CodeRabbit

* **New Features**
  * Added Call abstraction, unified TrackType names (VIDEO, SCREEN_SHARE) and participant-centric TTS/track handling.

* **Refactor**
  * Edge/Connection moved to async, event-driven model; call creation/join flows revised and audio track model simplified (AudioStreamTrack replaces queued track).
  * GetStream SFU event artifacts relocated into the GetStream plugin and import paths updated.

* **Documentation**
  * Updated protobuf generation and plugin import guidance.

* **Tests**
  * Expanded tests for multi-event registration and SFU event handling.

### DIFF
--- a/agents-core/vision_agents/core/agents/agents.py
+++ b/agents-core/vision_agents/core/agents/agents.py
@@ -198,7 +198,7 @@ class Agent:
 
         # Merge plugin events BEFORE subscribing to any events
         for plugin in [stt, tts, turn_detection, llm, edge, profiler]:
-            if plugin and hasattr(plugin, "events"):
+            if plugin is not None:
                 self.logger.debug(f"Register events from plugin {plugin}")
                 self.events.merge(plugin.events)
 


agents-core/vision_agents/core/agents/agents.py
Nash0x7E2: Some vibey code left in Agent it seems

agents-core/vision_agents/core/edge/call.py
```diff
@@ -0,0 +1,14 @@
+from typing import Protocol
+
+
+class Call(Protocol):
```
Nash0x7E2, line 4: Do you want to have a field here for custom metadata you can pass on to the underlying edge?

For example, Stream allows you to set a name for a call, other providers may allow you to do the same

dangusev: The metadata can already be passed to the `Edge.create_call` via `**kwargs`.
I'll change `Edge.create_call()` to make the return type generic, so users calling `edge.create_call` could receive a properly typed call object from the Edge

PR #348, opened 2026-02-09
Support multiple speakers on the same call

dangusev:
This PR adds a mechanism to pick the first active participant if there are multiple on the call.
This allows the Agent to process incoming audio as usual, even when speakers talk simultaneously. 

## What's changed
-  Added `id` field to `Participant`, which is supposed to be unique. The existing `user_id` field is not guaranteed be unique.
- `Agent` now buffers audio data per participant (previously, it was a single shared queue)
- Added `AudioFilter` interface and its `FirstSpeakerWinsFilter` implementation. It picks the first active speaker using Silero VAD, and processes only this audio track, ignoring others.  
The filtering is active only when there are 2+ participants on the call.

## Summary by CodeRabbit

* **New Features**
  * Multi-speaker handling: per-participant audio routing with a First-Speaker-Wins filter and parallel warmup of audio components.
  * Participant identity: Participant objects now include a stable id; agent speech events and TTS flows carry explicit participant info (replacing user_id).
  * STT behavior: STT integrations now require participant metadata for accurate attribution.

* **Tests**
  * Added comprehensive tests for the multi-speaker filter, speaker transitions, silence handling, and warmup behavior.

### DIFF
--- a/agents-core/vision_agents/core/agents/agents.py
+++ b/agents-core/vision_agents/core/agents/agents.py
@@ -984,6 +984,35 @@ class Agent:
         # Store the local video track.
         self._video_track_override_path = path
 
+    async def _poll_audio_queues(
+        self,
+        queues: dict[str, tuple[Participant, AudioQueue]],
+    ) -> AsyncIterator[tuple[Participant, PcmData]]:
+        """Poll all participant audio queues in parallel.
+
+        Yields (Participant, PcmData) tuples as each queue produces a chunk.
+        Queues that time out are silently skipped. On exit, any remaining
+        in-flight tasks are cancelled to avoid unhandled exceptions.
+        """
+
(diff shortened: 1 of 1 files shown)


agents-core/vision_agents/core/utils/audio_filter.py
```diff
@@ -0,0 +1,125 @@
...
+
+from getstream.video.rtc import PcmData
+from vision_agents.core.edge.types import Participant
+from vision_agents.core.vad.silero import SileroVADSession, SileroVADSessionPool
+from vision_agents.core.warmup import Warmable
+
+logger = logging.getLogger(__name__)
+
+
+class AudioFilter(abc.ABC):
```
Nash0x7E2, line 13: Nice

PR #501, opened 2026-04-21
Replace events as medium for passing data around the application

dangusev:
This PR introduces a major change by replacing `EventManager` and events as a way of
passing audio, transcripts and other data around the application with a new `Stream`
and `InferenceFlow` abstractions to manage components together.

The event bus was load-bearing for inference: STT → LLM → TTS handoffs, audio
chunks, transcripts, and turn boundaries all flowed as fan-out events. That made
backpressure impossible, ordering implicit, lifecycle leaks easy, and unit testing
of inference paths painful. With this change, each handoff is an explicit typed
`Stream`, and the orchestration that used to live in `Agent` is now a small set of
`InferenceFlow` implementations.

### New primitives

- `core/utils/stream.py` — `Stream[T]`: async iterator with backpressure (`maxsize`),
  `clear()` (drop buffered, keep iterators), `close()`, `peek()`, `collect(timeout=)`,
  and `send_iter()` for batch producers.
- `core/agents/inference/audio.py` — `AudioInputStream` / `AudioOutputStream`:
  PCM streams with chunk control messages and a flush path (final chunk now drains
  the carry-over buffer correctly).

### `InferenceFlow` abstraction

`Agent` no longer owns the STT/LLM/TTS state machine. It composes one of:

- `core/agents/inference/base.py` — abstract `InferenceFlow` (`start` / `stop` /
  `interrupt` / `simple_response` / `say` / `set_conversation`).
- `core/agents/inference/transcribing_flow.py` — STT → LLM → TTS pipeline with
  optional external `TurnDetector` (falls back to STT-driven turns; final empty
  transcripts no longer trigger turn end).
- `core/agents/inference/realtime_flow.py` — drives a `Realtime` LLM, consuming
  its new `Stream` outputs (`RealtimeAudioOutput`, `RealtimeUserTranscript`,
  `RealtimeAgentTranscript`, …) instead of events.

### `LLMTurn`

- `core/agents/inference/llm_turn.py` — owns a single LLM turn: response task,
  finalize task, cancellation, and a `Stream[LLMResponseDelta | LLMResponseFinal]`
  for downstream TTS. Cancels cleanly on interrupt; finalizes even when the LLM
  fails; logs background-task exceptions instead of swallowing them.

### LLM API

- `LLM.simple_response()` now returns an `AsyncIterator` of response deltas/finals
  rather than firing events.
- `simple_audio_response` docstring updated; OTel spans no longer include raw text.
- `FunctionRegistry.functions` is now public so `core.testing` can introspect tool
  calls without observing method calls.

### `Agent` slim-down

`agents-core/vision_agents/core/agents/agents.py`: **754 → 165 lines**. The
event-driven pipeline, turn handling, and audio plumbing moved into the
`InferenceFlow` implementations. Audio-producer task only starts when the agent
actually publishes audio.

### Metrics

- `core/observability/collector.py` — `MetricsCollector` no longer subscribes to
  events; plugins call `_emit_*` helpers directly. `merge()` is idempotent.
- `EdgeTransport` constructs a `MetricsCollector` by default.
- Base `TTS` now receives a `MetricsCollector`.

### Transcripts

- `TranscriptBuffer.final` property to detect a finalized buffer.
- `TranscriptStore` methods accept `drop: bool` so callers can accumulate finals
  without deleting them.

### Misc

- `core/utils/tokenizer.py` — `TTSTokenizer` sanitizes the final chunk too.
- `core/utils/exceptions.py` — `log_task_exception` / `log_exceptions` used across
  background tasks; does not catch `BaseException`.
- `core/stt/stt.py` — `Transcript` uses `self.response` as its source of truth.
- `core/tts/manual_test.py` and the old `core/tts/testing.py` helpers removed.

## Plugin updates

All plugins are migrated to the new APIs. No behavior change beyond what the new
approach implies.

## Tests / cleanup

- Removed old event-driven `Agent` tests; new tests cover the inference flows.
- `core.testing` updated: `collect_simple_response` lives here; sessions assert on
  streams/state, not on observed method calls.
- `pytest-cov` added to dev requirements.
- Examples (`nemotron_example.py`, others) and `agent.llm.simple_response()` call
  sites updated for the new iterator API.

## Broken things & TODOs

To limit the scope of this already huge PR, some things are left for the follow-up PRs:

- Avatar plugins need to be updated to use the new API - **I'll create a PR to this one.**
- AudioProcessors do not work anymore.
- `docs/ai/***` and `DEVELOPMENT.md` need updating


agents-core/vision_agents/core/agents/inference/transcribing_flow.py
```diff
@@ -0,0 +1,514 @@
...
+              regardless of their transcript.
+              They may be received long before the final transcript is avaialble.
+
+
+        """
+        async for event in stt_output:
+            with log_exceptions(logger, "Error while processing STT output"):
+                participant = event.participant
+                if self._llm_turn and self._llm_turn.finalized:
+                    self._llm_turn = None
```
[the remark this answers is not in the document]
dangusev: llmturn.finalized is true only after both tasks are done, which means they cannot be cancelled at this point

plugins/aws/vision_agents/plugins/aws/aws_llm.py
```diff
@@ -97,636 +103,390 @@ def _create_client():
...
-        """
-        Converse gives full access to the Bedrock Converse API.
-        This method wraps the Bedrock method and broadcasts events for agent integration.
-        """
-        if "modelId" not in kwargs:
-            kwargs["modelId"] = self.model
+        if participant is None:
+            await self._conversation.send_message(
+                role="user", user_id="user", content=text
+            )
```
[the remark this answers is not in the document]
dangusev: It's a known issue, will be fixed separately

agents-core/vision_agents/core/tts/tts.py
```diff
@@ -300,19 +297,132 @@ async def send(
...
+                    real_time_factor=real_time_factor,
                 )
             )
+            self.metrics.on_tts_synthesis(
+                provider=self.provider_name,
+                synthesis_time_ms=synthesis_time * 1000,
+                audio_duration_ms=estimated_audio_duration_ms,
+                character_count=len(text),
+            )
+        except Exception:
```
aliev: This try/except looks redundant.

agents-core/vision_agents/core/turn_detection/turn_detection.py
```diff
@@ -29,44 +46,25 @@ def __init__(
         self.provider_name = provider_name or self.__class__.__name__
         self.events = EventManager()
         self.events.register_events_from_module(events, ignore_not_compatible=True)
+        self.metrics = MetricsCollector()
+        self._output: Stream[TurnEnded | TurnStarted] = Stream()
```
aliev, line 50: nit: this union is already repeated (see also: stt.py) and represents a stable domain concept (turn lifecycle events). Let’s introduce a dedicated type alias (e.g. TurnEvent) and use it across the pipeline.

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -415,210 +353,94 @@ async def on_call_ended(_: CallEndedEvent):
...
+        else:
+            return TranscribingInferenceFlow(
+                audio_input=self._audio_input_stream,
+                audio_output=self._audio_output_stream,
+                stt=self.stt,
+                llm=self.llm,
+                tts=self.tts,
+                turn_detector=self.turn_detection,
+                conversation=self.conversation
+                or InMemoryConversation(
```
aliev, line 434: nit: consider extracting shared arguments into a common_kwargs dict to reduce duplication

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -234,22 +230,26 @@ def __init__(
...
+        self._collector = MetricsCollector()
+        # Merge plugin metric collectors so plugin on_*() calls forward to the agent root
+        for plugin in [stt, tts, turn_detection, llm, *self.processors]:
             if plugin is not None:
-                self.logger.debug(f"Register events from plugin {plugin}")
-                self.events.merge(plugin.events)
+                self._collector.merge(plugin.metrics)
+
+        # Merge plugin events BEFORE subscribing to any events
+        for component in [stt, tts, turn_detection, llm, edge, profiler]:
```
aliev, line 238: The `stt` and `turn_detection` can be removed, since they don't emit anything through `.events`

agents-core/vision_agents/core/stt/stt.py
```diff
@@ -38,129 +81,100 @@ def __init__(
         self.provider_name = provider_name or self.__class__.__name__
 
         self.events = EventManager()
```
aliev, line 83: This EventManager is a no-op, nothing writes to it and no one subscribes. Worth removing along with `core/stt/events.py`

agents-core/vision_agents/core/turn_detection/turn_detection.py
```diff
@@ -29,44 +46,25 @@ def __init__(
         self.provider_name = provider_name or self.__class__.__name__
         self.events = EventManager()
```
aliev, line 47: Same story as STT: this EventManager is a no-op, nothing writes to it. Worth removing along with `core/turn_detection/events.py` (the only consumer is `examples/11_moderation_example`, which should move to the Stream API)

agents-core/vision_agents/core/tts/tts.py
```diff
@@ -201,6 +197,7 @@ async def stop_audio(self) -> None:
         """
         pass
 
+    # TODO: This will be gone
     async def send(
```
aliev: This is unused now (all callers moved to send_iter), the docs are stale too

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -992,97 +809,6 @@ def _on_rtc_reconnect(self):
         # TODO: implement me
```
aliev: Might be worth removing this while we're already touching this area

agents-core/vision_agents/core/agents/inference/__init__.py
```diff
@@ -0,0 +1,20 @@
+from .audio import (
```
aliev, line 1: nit: we use `__all__` for explicit re-exports elsewhere in core/ (see `stt/__init__.py`). worth using the same pattern

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -1134,6 +860,11 @@ async def _consume_incoming_audio(self) -> None:
                             if pcm is None:
                                 continue
 
+                        await self._audio_input_stream.send(
+                            AudioInputChunk(data=pcm, participant=participant)
+                        )
+                        continue
+
                         # Pass PCM through audio processors
```
aliev, line 856: Looks like a leftover from the refactor. The continue above makes this block unreachable

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -1163,6 +889,25 @@ async def _consume_incoming_audio(self) -> None:
...
+    async def _produce_audio_output(self):
+        if self._audio_track is None:
+            return
+
+        try:
+            while self._call_ended_event and not self._call_ended_event.is_set():
+                async for audio_output in self._audio_output_stream:
+                    match audio_output:
+                        case AudioOutputChunk(data=data) if data is not None:
+                            await self._audio_track.write(audio_output.data)
```
aliev: The guard narrows data, but the body re-reads the attribute via audio_output.data, which the type checker can't assume is the same value (could be a property):

```
Argument to bound method `AudioStreamTrack.write` is incorrect: Expected `PcmData`, found `PcmData | None`
```

use the captured data directly:

```suggestion
                            await self._audio_track.write(data)
```

agents-core/vision_agents/core/agents/agents.py
```diff
@@ -1441,33 +1150,13 @@ def _validate_configuration(self):
                     "At least one processing capability (audio or video) is required"
                 )
 
-    def _prepare_rtc(self):
+    def _setup_media_tracks(self):
         if self.publish_audio:
             if self.audio_publishers:
                 self._audio_track = self.audio_publishers[0].publish_audio_track()
```
aliev, line 1135: My type checker flags this line:

```
Object of type `AudioStreamTrack` is not assignable to attribute `_audio_track` of type `AudioStreamTrack | None`
```

`aiortc.AudioStreamTrack` and `getstream.video.rtc.AudioStreamTrack` are sibling classes. Both inherit from `aiortc.mediastreams.MediaStreamTrack` directly, so neither is assignable to the other.

On top of that, getstream's variant has `write()` / `flush()` and aiortc's doesn't.

All concrete implementations (lemonslice, anam, heygen) actually return the getstream variant, and `Agent._audio_track` is typed as the getstream one too — so the abstract signature here is the only place still pointing at aiortc. Looks like an oversight from when the protocol was introduced.

Suggestion: in `agents-core/vision_agents/core/processors/base_processor.py:103`, replace:

```python
def publish_audio_track(self) -> aiortc.AudioStreamTrack: ...
```

with:

```python
def publish_audio_track(self) -> AudioStreamTrack: ...
```

This will need `from getstream.video.rtc import AudioStreamTrack` at the top of the file. The existing `import aiortc` should stay — it's still used for `VideoStreamTrack`, where there's no collision and the type is correct.

aliev (changes requested):

agents-core/vision_agents/core/agents/agents.py
dangusev: applied locally 👍

agents-core/vision_agents/core/tts/tts.py
dangusev: forgot to emit tts error metrics, fixing

agents-core/vision_agents/core/agents/agents.py
dangusev: I prefer kwargs expanded this way to keep typing info

agents-core/vision_agents/core/agents/agents.py
dangusev: I keep `events` to let other implementations define their own

agents-core/vision_agents/core/stt/stt.py
dangusev: same logic here: we still need events, and I plan to retro-fit some of them back in the upcoming PRs

agents-core/vision_agents/core/tts/tts.py
dangusev: will remove

agents-core/vision_agents/core/agents/agents.py
dangusev: done

aliev (approved):

PR #520, opened 2026-04-27
Anam plugin optimizations

DaemonLoki:
## Why

- Optimize Anam plugin
- We don't need to chunk the TTS audio, since Anam can handle arbitrary TTS size
- No resampling needed

## Changes

- add option for setting the audio chunk size manually (in initializer of `anam_avatar_publisher.py`)
- remove resampling of audio data


dangusev:
The changes from this PR will be added as a part of https://github.com/GetStream/Vision-Agents/pull/534.

Cheers.

PR #529, opened 2026-05-01
fix(mcp): own MCP transport lifecycle in a single supervisor task

aliev:
## Why

`MCPServerRemote.connect()` and `disconnect()` drive `streamablehttp_client` and `ClientSession` contexts directly. Those contexts are built on anyio cancel scopes that require `__aenter__` and `__aexit__` to run in the same asyncio task. The agent's `_close()` runs as a shielded task, so the disconnect path executes in a different task than connect. That trips anyio's cross-task cancel-scope check, leaves a half-cancelled scope attached to the running event loop, and `_deliver_cancellation` reschedules itself indefinitely via `call_soon`, pegging a CPU core near 100% for the remaining lifetime of the process.

A `py-spy dump` of an affected process confirms the spinning frame is `anyio._backends._asyncio.CancelScope._deliver_cancellation`. The agent logs the underlying warning on every session close: `Error closing MCP session context: Attempted to exit cancel scope in a different task than it was entered in`.

This is a known issue class in `modelcontextprotocol/python-sdk` with no near-term fix planned ([note on #922](https://github.com/modelcontextprotocol/python-sdk/issues/922#issuecomment-3338041354)). modelcontextprotocol/python-sdk#1805 describes the same CPU-after-disconnect symptom on the streamable-http transport, and modelcontextprotocol/python-sdk#577 tracks the cross-task cleanup failure directly. The fix therefore belongs at the VA integration layer.

## Changes

A supervisor task in `MCPServerRemote` now enters and exits both transport contexts via `AsyncExitStack`. `connect()` spawns the supervisor and awaits a setup future the supervisor resolves on success or rejects on failure; `disconnect()` signals stop via an event and awaits the supervisor's natural exit. The transport contexts therefore always enter and exit on the same asyncio task.

`tests/test_mcp_server_remote_lifecycle.py` pins the invariant under cross-task `connect()` / `disconnect()` and covers the existing already-connected behaviour (no-op plus warning, no raise).


aliev:
@coderabbitai review

agents-core/vision_agents/core/mcp/mcp_server_remote.py
```diff
@@ -47,97 +67,90 @@ async def connect(self) -> None:
...
-            self.logger.error(f"Failed to connect to remote MCP server: {e}")
-            # Clean up any partial connection state
-            await self._cleanup_connection()
+            await self._setup
+        except (Exception, asyncio.CancelledError):
+            # Setup failed, or connect() itself was cancelled. Either way,
+            # tear the supervisor down and let its cleanup finish before
+            # propagating. Anything else (SystemExit, KeyboardInterrupt) is
+            # left to propagate immediately without our cleanup detour.
+            await self._teardown_supervisor()
```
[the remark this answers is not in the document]
aliev: Logger fixes applied in 7aab7d9 for L127-133 and L139-140, and in ffdbf2d for L151-154 (info-level caveat retained, traceback now emitted at DEBUG via `exc_info=True` so operators can opt in without polluting normal logs)

L78-83 catches only to run `_teardown_supervisor()` and re-raise. The type filter doesn't change handling, only which classes trigger teardown

tests/test_mcp_server_remote.py
```diff
@@ -0,0 +1,154 @@
...
+These tests pin the contract: regardless of which task drives ``connect()``
+and ``disconnect()``, the underlying transports' ``__aenter__`` / ``__aexit__``
+must run on the same asyncio task.
+"""
+
+import asyncio
+import logging
+from typing import Any, Optional
+from unittest.mock import AsyncMock
+
```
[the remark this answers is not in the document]
aliev: Observing the task identity of `__aenter__` / `__aexit__` calls into an external library requires instrumenting that boundary, which is what these mocks are for

agents-core/vision_agents/core/mcp/mcp_server_remote.py
```diff
@@ -47,97 +65,93 @@ async def connect(self) -> None:
...
-            # Clean up the connection
-            await self._cleanup_connection()
-
-            self._is_connected = False
-            self.logger.info("Disconnected from remote MCP server")
-
+            if self._stop_event is not None:
+                self._stop_event.set()
+            if self._supervisor_task is not None:
+                await self._supervisor_task
```
dangusev: Why not simply cancel the task?
There's a `cancel_and_wait` helper for that

tests/test_mcp_server_remote.py
```diff
@@ -0,0 +1,154 @@
...
+    assert "Already connected" in caplog.text
+
+    async def cleanup() -> None:
+        await server.disconnect()
+
+    await asyncio.create_task(cleanup())
+
+
+@pytest.mark.asyncio
```
dangusev: not needed anymore

tests/test_mcp_server_remote.py
dangusev, line 1: Let's name it simply `test_mcp_server_remote.py`

dangusev (approved):

PR #534, opened 2026-05-06
Add dedicated Avatar plugin type

dangusev:
## Why
Previously, avatar plugins were implemented as a combination of AudioProcessor and VideoProcessor.

As a result, their video tracks were mixed with other video processors, and the agent wasn't sending the correct track to VideoLLMs and Realtimes.
With an avatar on the call, the Agent was sending the avatar's video to the model instead of the participant's track.

It was also a bit confusing because it was possible to pass multiple processors, but multiple avatars for a single agent make no sense.

## Changes
-  Added a new dedicated `Avatar` and the `avatar` parameter to the Agent class.  
When an avatar is passed, it replaces the agent's output video track with its own, streaming video to the call.

- Converted existing avatar plugins to the new API.
  -  Also, added a new `liveavatar` plugin to replace the outdated `heygen` one.   
  `heygen` plugin will be deprecated in this PR and removed in future releases.

## Example API
```python
# New API
agent = Agent(avatar=Avatar(), ...)
```


dangusev:
@coderabbit review

aliev (approved):

PR #578, opened 2026-05-21
perf(agents-core): parallelize component lifecycle via shared Component base

aliev:
## Why

`Agent._apply` was awaiting `start` / `stop` / `close` on each component sequentially, so the slow ones (RTC/WebSocket connects in STT, TTS, avatar plugins) stacked on top of each other. On a realistic setup — Gemini LLM + ElevenLabs STT + Deepgram TTS + vogent + LemonSlice avatar — this cost **~4.3 s** in `_apply("start")` alone, and the user felt every second of it as "time to first reply".

Components in `start`/`close` don't depend on one another (verified by walking the base classes and plugin implementations), so they can run concurrently. After parallelizing, agent startup is almost twice as fast:

| run | `_apply("start")` |
|---|---|
| sequential (before) | 4295 ms |
| concurrent (this PR) | 2354 ms |

That's ~1.9 s shaved off agent join, dominated now by the single slowest component (LemonSlice RTC connect).

### Asymmetric error handling

Starting and stopping have different correctness needs and are now exposed as two distinct methods:

- **`_start_components`** — if a component cannot start (e.g. STT websocket connect fails), the agent must abort `join()` instead of running half-initialised. The first failure is logged and re-raised; in-flight siblings are explicitly cancelled and awaited so we don't leak orphan network connects (default `asyncio.gather` does **not** cancel siblings on first exception).
- **`_stop_components`** — best-effort: one broken plugin's `close` must not prevent others from releasing resources. Errors are logged via `log_exceptions` and swallowed.

### Unified `Component` base class

The old `_apply(function_name)` only existed because component classes declared lifecycle hooks ad-hoc — some had `start`, some didn't, `TurnDetector` used `stop` instead of `close`. We unify them under a single `Component` base class (`agents-core/vision_agents/core/base.py`) with no-op `start` and `close` defaults, so the agent can call them directly without ``getattr``-based dispatch.

Shape change: `TurnDetector.stop` is renamed to `close` (base class + vogent and smart_turn plugins, their tests, and the smart_turn README). `Processor` and `EdgeTransport` drop their redundant abstract close declarations; the contract now comes from `Component`.

### Tests

Four new unit tests in `TestAgent`:
- parallel timing of `_start_components` (catches accidental reversion to sequential)
- `agent.join()` propagates a component `start` failure (catches removal of the fail-fast call)
- `_stop_components` swallows a failing component (catches accidental flip of the semantic)
- sibling cancellation when one `start` raises (catches removal of the cancel-and-await cleanup)

## Changes

- New `Component` base class in `agents-core/vision_agents/core/base.py`. `LLM`, `STT`, `TTS`, `TurnDetector`, `EdgeTransport`, `Avatar`, `Processor` now inherit from it.
- `TurnDetector.stop` renamed to `close` (base + vogent + smart_turn + tests + README).
- `Agent._apply` and `_safe_invoke` replaced by `_start_components` (fail-fast + sibling cancellation) and `_stop_components` (best-effort).
- `agent.join()` calls `_start_components`; `agent._close()` calls `_stop_components` (collapsing the old back-to-back `_apply("stop") + _apply("close")`).
- `processors` docstring simplified: frame dispatch in list order, lifecycle hooks concurrent.
- `log_exceptions` widened to accept `logging.LoggerAdapter` so it works with `_AgentLoggerAdapter`.


aliev:
@coderabbitai review

dangusev (approved):

aliev:
@coderabbitai review

PR #580, opened 2026-05-22
test(agents-core): add unit tests for log_exceptions

aliev:
## Why

`log_exceptions` is used in eight places across the codebase but has no direct test coverage. Callers (in particular the lifecycle error handling in `Agent._apply`) rely on a specific contract — silently changing what gets caught would break those callers without anything failing.

Pin down the contract before it can drift.

## Changes

New `tests/test_utils/test_exceptions.py` covering:

- catches `Exception` by default and logs via `logger.exception`
- `reraise=True` propagates after logging
- only catches the configured exception tuple; others propagate
- `BaseException` subclasses outside `Exception` (`KeyboardInterrupt`, `asyncio.CancelledError`) pass through unchanged — load-bearing for cancellation flows

### DIFF
--- a/tests/test_utils/test_exceptions.py
+++ b/tests/test_utils/test_exceptions.py
@@ -6,11 +6,13 @@ import pytest
 from vision_agents.core.utils.exceptions import log_exceptions
 
 
-logger = logging.getLogger("test_log_exceptions")
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


aliev:
@coderabbitai review

dangusev (approved):

PR #581, opened 2026-05-23
feat(cli/init): scaffold tests/Dockerfile and simplify run command

aliev:
## Why

`vision-agents init` produced a minimal project that left two onboarding gaps:

1. **No test scaffolding.** Nothing pointed new users at `vision_agents.testing`, so the first time most agents got exercised was inside a live call.
2. **`uv run vision-agents agent run` was the suggested invocation.** It works, but it adds an extra concept (the CLI dispatches to `[tool.vision-agents.agent].entrypoint`) right when the user is trying to understand what `agent.py` *is*.

This PR ships a small, opinionated set of files so a freshly-generated project is something you can run, test, and deploy without leaving the directory.

## Changes

- Scaffold a `tests/` directory with an example built on `vision_agents.testing` (`TestSession`, `LLMJudge`, multi-turn). `pytest` is configured under `[tool.pytest.ini_options]` in `pyproject.toml`; `pytest` + `pytest-asyncio` land in a `[dependency-groups] dev` group.
- Expose `INSTRUCTIONS`, `MODEL`, and `create_llm()` from `agent.py` so tests import the same setup as production. Production keeps `gemini.Realtime` for audio; `create_llm` is the text-mode factory tests use.
- Add a `Dockerfile` (python:3.12-slim + `uv`, layered for cache, runs `serve --host 0.0.0.0`) and a `.dockerignore`.
- Switch the README and the init "next steps" output from `uv run vision-agents agent run` to `uv run agent.py run` — same behavior, but the command now mirrors what's in `agent.py` (`if __name__ == "__main__": runner.cli()`). Django's `python manage.py` was the inspiration.
- Teach the scaffolder to render nested template paths so `tests/` works.

## Summary by CodeRabbit

* **New Features**
  * Docker containerization support added for scaffolded projects with Python 3.12 slim containers and two-stage dependency caching.
  * Test infrastructure and examples included in generated projects using pytest and LLM-based evaluation.

* **Documentation**
  * Updated CLI command syntax examples in generated templates and README.
  * Added Docker ignore configuration to newly scaffolded projects.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/Vision-Agents/pull/581?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/agents-core/vision_agents/cli/init/templates/agent.py.j2
+++ b/agents-core/vision_agents/cli/init/templates/agent.py.j2
@@ -11,13 +11,6 @@ INSTRUCTIONS = (
     "Keep responses short and conversational."
 )
 
-MODEL = "gemini-2.5-flash"
-
-
-def create_llm() -> gemini.LLM:
-    """Text `gemini.LLM` used by tests (production uses `gemini.Realtime`)."""
-    return gemini.LLM(MODEL)
-
 
 async def create_agent(**kwargs) -> Agent:
     return Agent(
--- a/agents-core/vision_agents/cli/init/templates/agent.py.j2
+++ b/agents-core/vision_agents/cli/init/templates/agent.py.j2
@@ -6,14 +6,24 @@ from vision_agents.plugins import getstream, gemini
 load_dotenv()
 
 
+INSTRUCTIONS = (
+    "You're a helpful voice AI assistant. "
+    "Keep responses short and conversational."
+)
+
+MODEL = "gemini-2.5-flash"
+
+
+def create_llm() -> gemini.LLM:
+    """Text `gemini.LLM` used by tests (production uses `gemini.Realtime`)."""
+    return gemini.LLM(MODEL)
+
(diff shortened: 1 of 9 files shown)


dangusev (approved):

agents-core/vision_agents/cli/init/templates/agent.py.j2
```diff
@@ -6,14 +6,24 @@ from vision_agents.plugins import getstream, gemini
...
 
+INSTRUCTIONS = (
+    "You're a helpful voice AI assistant. "
+    "Keep responses short and conversational."
+)
+
+MODEL = "gemini-2.5-flash"
+
+
+def create_llm() -> gemini.LLM:
```
dangusev: Let's return the generic `core.LLM` here instead of `gemini.LLM`. Makes experimenting with different models a little bit easier.

PR #593, opened 2026-06-03
feat: add MiniMax LLM provider plugin with M3 as default

octo-patch:
## Summary

Add a new MiniMax LLM provider plugin to Vision Agents. The plugin uses MiniMax's OpenAI-compatible Chat Completions API at `https://api.minimax.io/v1` and ships with **MiniMax-M3 as the default model** (512K context, up to 128K output, image input support).

## Changes

- New plugin package under `plugins/minimax/`:
  - `vision_agents/plugins/minimax/minimax_llm.py` — `MiniMaxLLM` implementation
  - `vision_agents/plugins/minimax/__init__.py` — exports `LLM`
  - `README.md`, `pyproject.toml`, `py.typed`
  - `tests/test_minimax_llm.py` — unit + integration tests
- Wire the plugin into the uv workspace (`pyproject.toml`) and the `agents-core` extras list (`minimax = ["vision-agents-plugins-minimax"]`).
- Add `MINIMAX_API_KEY` and `MINIMAX_BASE_URL` entries to `.env.example`.
- Mention MiniMax in the README LLM integrations list.

## Supported models

| Model | Notes |
|-------|-------|
| `MiniMax-M3` (default) | Latest flagship. 512K context, 128K max output, image input. |
| `MiniMax-M2.7` | Previous generation flagship. |
| `MiniMax-M2.7-highspeed` | Low-latency variant of M2.7. |

## Usage

```python
from vision_agents.plugins import minimax

llm = minimax.LLM()  # defaults to MiniMax-M3
llm = minimax.LLM(model="MiniMax-M2.7")
```

## Testing

- Unit tests pass (`uv run --no-sync pytest plugins/minimax/tests/ -m "not integration"` → 6 passed).
- Lint and format are clean (`uv run --no-sync ruff check plugins/minimax/`, `ruff format --check`).
- Integration tests are skipped in CI when `MINIMAX_API_KEY` is not set.

## Notes

- Default base URL is `https://api.minimax.io/v1` (overseas). Do not use `api.minimax.chat` for production traffic.
- Default `temperature` is `1.0` (the API rejects `0.0`).
- The `response_format` field is intentionally not exposed; the MiniMax Chat Completions API does not support it.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,3 +32,7 @@ ROBOFLOW_API_URL=your_roboflow_api_url_here
 
 # LemonSlice API credentials
 LEMONSLICE_API_KEY=your_lemonslice_api_key_here
+
+# MiniMax API credentials
+MINIMAX_API_KEY=your_minimax_api_key_here
+MINIMAX_BASE_URL=https://api.minimax.io/v1
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ agent = Agent(
 
 ## Out-of-the-Box Integrations
 
-**LLMs:** [OpenAI](https://visionagents.ai/integrations/openai) · [Gemini](https://visionagents.ai/integrations/gemini) · [xAI](https://visionagents.ai/integrations/xai) · [OpenRouter](https://visionagents.ai/integrations/openrouter) · [Hugging Face](https://visionagents.ai/integrations/huggingface) · [Kimi AI](https://visionagents.ai/integrations/kimi)
+**LLMs:** [OpenAI](https://visionagents.ai/integrations/openai) · [Gemini](https://visionagents.ai/integrations/gemini) · [xAI](https://visionagents.ai/integrations/xai) · [OpenRouter](https://visionagents.ai/integrations/openrouter) · [Hugging Face](https://visionagents.ai/integrations/huggingface) · [Kimi AI](https://visionagents.ai/integrations/kimi) · [MiniMax](https://visionagents.ai/integrations/minimax)
 
 **Realtime:** [OpenAI Realtime](https://visionagents.ai/integrations/openai) · [Gemini Live](https://visionagents.ai/integrations/gemini) · [AWS Nova Sonic](https://visionagents.ai/integrations/aws-bedrock) · [Qwen](https://visionagents.ai/integrations/qwen) · [Inworld](https://visionagents.ai/integrations/inworld)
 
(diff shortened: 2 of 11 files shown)


octo-patch:
Thanks @coderabbitai for the careful review. Addressed in commit 7f8deb5; reasoning per item below.

**Accepted and fixed:**

1. **`tools_max_rounds` silently rewritten** — now raises `ValueError("tools_max_rounds must be >= 1, got <n>")` instead of `max(tools_max_rounds, 1)`. Matches the agents-core constructor-validation guidance.
2. **Closing injected `AsyncOpenAI` clients** — added `self._owns_client` flag; `close()` only tears down clients we constructed internally. Injected clients now stay alive for their owner.
3. **Alphabetical ordering in `pyproject.toml`** — moved `vision-agents-plugins-minimax` between `mistral` and `assemblyai` in both `[tool.uv.sources]` and workspace members.

**Deferred, with reasoning:**

4. **Per-instance `_pending_tool_calls` dict (lines 75 / 210 / 428)** — the dict is reset at each request entry, and current `MiniMaxLLM` usage in the framework is single-flow-per-instance (one streaming call active at a time, like the other LLM plugins here). Migrating to a local-state-passed-through-helpers model is a non-trivial refactor with no current concurrent-usage trigger. Will follow up separately if/when concurrent `simple_response` on a single instance becomes a supported pattern.
5. **Skip `participant is None` branch (lines 114-117)** — this mirrors the existing convention in this repo: the STT path appends the user message to the conversation *before* `simple_response()` is invoked with a participant, so we deliberately skip the duplicate append in that branch. Other plugins (e.g. anthropic) tolerate the same pattern via message-merge logic on the request side. Changing this would risk double-appending the STT utterance.
6. **Integration `skipif` for `MINIMAX_API_KEY`** — kept as `skipif` to match the existing repo convention used by other plugin integration suites that gate on provider secrets; a hard fail here would diverge from how the rest of the matrix opts into integration runs.
7. **`test_default_temperature_is_one` mocking** — this is a small unit boundary specifically around the API-rejects-0 contract (MiniMax returns 400 if `temperature=0`). The full behavior contract is covered in the integration suite (`TestMiniMaxLLMIntegration.test_simple_response`). I can drop the unit if you prefer to lean only on the integration test.
8. **M3 context-window doc in `plugins/minimax/README.md`** — the README currently says "512K context." Per MiniMax's official docs, M3 supports up to 1M input tokens and pricing is tiered at `<=512K` vs `>512K`. The plugin itself doesn't register pricing (that's LiteLLM's job downstream), so we're not exposing a stale numeric constant in code. Happy to update the README to say something like "M3 (up to 1M input tokens, pricing tiered at 512K)" if that's more useful — let me know which framing you prefer.

Verification: `uv run --no-sync pytest plugins/minimax/tests/test_minimax_llm.py -m "not integration"` → 6/6 passing; `uv run --no-sync ruff check plugins/minimax/` → clean.

octo-patch:
Pushed a ruff format fix for the formatter check. The remaining red Test job is unrelated to this plugin — it's the gemini suite erroring on a missing Google API key in CI, not the minimax tests (those pass).

On the M3 context window note: we're intentionally documenting 512K to match what's currently exposed through the OpenAI-compatible endpoint this plugin uses, rather than the marketing 1M figure. Happy to revisit if you'd prefer the headline number.

PR #607, opened 2026-06-25
Add TwelveLabs Pegasus video understanding plugin

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## Why

Vision Agents already supports several frame-by-frame VLMs (Moondream, NVIDIA, etc.) for the "any model" vision angle. TwelveLabs **Pegasus** is a video-understanding model: instead of reasoning over a single still frame, it analyzes a short video clip, so an agent can answer "what just happened?" with an understanding of motion and events over time. This PR adds Pegasus as a first-class `VideoLLM` so it can be dropped into an `Agent` exactly like the existing vision models.

## Changes

- New opt-in package `plugins/twelvelabs` (`vision-agents-plugins-twelvelabs`) exposing `PegasusVLM`, a `llm.VideoLLM` subclass. It buffers recent frames from the watched track, encodes a short MP4 clip on demand, uploads it to the TwelveLabs Assets API, and streams the analysis back (vocalizable by the agent's TTS).
- Mirrors the existing Moondream cloud-VLM conventions: `watch_video_track` (own or shared `VideoForwarder`), streaming `simple_response` yielding `LLMResponseDelta`/`LLMResponseFinal`, metrics, and async `close()`.
- Wired into the uv workspace and the `agents-core` optional-dependency extras the same way as every other plugin (`vision-agents[twelvelabs]`). No defaults changed; nothing existing is touched.
- Tests: no-network unit tests (constructor validation, frame-buffer sizing, and a real MP4 encode round-trip asserting the 360x360 minimum-resolution upscale is decodable) plus an integration test gated on `TWELVELABS_API_KEY` (skips without it).
- Docs: plugin `README.md` with install, quick-start, parameters, and the Pegasus 360x360 / 4-second-minimum constraints.

### Testing

- `uv run --no-sync ruff check` / `ruff format --check` / `mypy` — clean on the new files.
- `uv run --no-sync pytest plugins/twelvelabs/tests -m "not integration"` — 5 passed.
- `uv run --no-sync pytest plugins/twelvelabs/tests -m integration` with a real key — passed end-to-end: clip uploaded (Assets API `201`), `/analyze` returned `200`, and Pegasus streamed a correct description of the buffered clip.

This is opt-in and non-breaking. You can grab a free API key at https://twelvelabs.io — there's a generous free tier.


Nash0x7E2:
Hi @mohit-twelvelabs - Thank you for your contribution, we're reviewing the PR. Looks like Coderabbit has some notes, can you take a look?

mohit-twelvelabs:
Thanks for the review @Nash0x7E2! Worked through CodeRabbit's notes in d226b73c:

1. **Integration test skip** — Dropped the `skipif` on `TestPegasusVLMIntegration`. The opt-in test now asserts the key is present with an actionable message (and `PegasusVLM()` itself already raises a clear missing-key error), so explicitly-selected integration runs fail loudly instead of silently skipping.

2. **Min clip duration + FPS precision** — `_encode_clip` now encodes with `fractions.Fraction(self.fps).limit_denominator(1000)` instead of `int(self.fps)`, so non-integer rates like 29.97 are preserved exactly. Frame counts use `math.ceil`, and the clip is padded (repeating the last frame) up to the 4s Pegasus floor. Verified locally: at 29.97 fps the encoded clip is 4.004s at rate 29.97 (previously `int(29.97)=29` lost the rate and `int(fps*secs)` truncated to 3.97s).

3. **Forwarder handler leak** — When swapping to a `shared_forwarder` we now `remove_frame_handler(self._on_frame_received)` from the previous forwarder first. `stop_watching_video_track` also detaches the handler instead of calling `stop()` outright, matching the sibling plugins (gemini/nvidia/etc.) and leaving a shared forwarder alive for other subscribers.

4. **Asset cleanup** — Added a best-effort `client.assets.delete(asset_id, force=True)` in a `finally` block after analysis, so uploaded clips don't persist. Cleanup failures are logged via `logger.exception` and never mask the analysis result/error.

5. **Response logging** — Replaced `logger.info(f"Pegasus response: {total_text}")` with a debug-level length log (`Pegasus response received (%d chars)`) so generated text isn't logged.

Also added the missing `__init__ -> None` annotation. `ruff check` / `ruff format --check` clean, `mypy` clean on the changed file, and the 5 unit tests pass (`pytest -m "not integration"`).

mohit-twelvelabs:
Cleared the remaining CodeRabbit note too (markdownlint MD040 — added a language specifier to the `.env` fenced block in the example README) in 3557e64. CodeRabbit should be fully green now.

Nash0x7E2 (approved):
@mohit-twelvelabs - Added a basic example to the plugin folder so developers wanting to try the model has an easy entry path.

mohit-twelvelabs:
Thank you @Nash0x7E2 — and thanks for adding the example to smooth the entry path! 🎉 Great to have TwelveLabs Pegasus in Vision-Agents. Happy to help with any follow-ups (e.g. Marengo embeddings as a video search/retrieval option) if that'd be useful down the line.

Nash0x7E2:
Awesome @mohit-twelvelabs, let's chat more over email

PR #612, opened 2026-07-14
Support the new stateful AudioStreamTrack from `getstream`

dangusev:
## Why
Stateful resampler provides better audio quality with fewer hisses and artifacts.

## Changes
- pass final=True to `audio_track.write()` to flush the resampler
- use stateful resampling in Avatars, LocalEdge and TencentEdge
- extra: add close timeout for Anam to prevent hanging when the call ends
- Updated `getstream` requirement to v4.1.0+

### DIFF
--- a/plugins/anam/vision_agents/plugins/anam/anam_avatar.py
+++ b/plugins/anam/vision_agents/plugins/anam/anam_avatar.py
@@ -249,7 +249,9 @@ class AnamAvatar(Avatar):
             return
         # Flush the resampler tail so the last partial frame isn't dropped.
         for frame in self._resampler.flush():
-            await self._audio_input_stream.send_audio_chunk(frame.to_ndarray().tobytes())
+            await self._audio_input_stream.send_audio_chunk(
+                frame.to_ndarray().tobytes()
+            )
         await self._audio_input_stream.end_sequence()
 
     async def _connect(self) -> None:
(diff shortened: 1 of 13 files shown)


plugins/anam/vision_agents/plugins/anam/anam_avatar.py
```diff
@@ -116,6 +121,9 @@ def __init__(
         self._exit_stack = contextlib.AsyncExitStack()
         self._real_session: Session | None = None
         self._audio_input_stream: AgentAudioInputStream | None = None
+        self._resampler = FrameResampler(
+            rate=AVATAR_SAMPLE_RATE, layout="mono", format="s16", frame_size=0
```
dangusev, line 125: why `frame_size=0` ?

plugins/tencent/Dockerfile
```diff
@@ -2,8 +2,7 @@ FROM python:3.12-slim
 
 # Pick up Debian security patches that aren't baked into the base tag yet.
 RUN apt-get update \
-    && apt-get upgrade -y \
-    && rm -rf /var/lib/apt/lists/*
+    && apt-get upgrade -y && apt-get install -y git && rm -rf /var/lib/apt/lists/*
```
dangusev: remove `git` here

plugins/anam/vision_agents/plugins/anam/anam_avatar.py
dangusev: Because we don't need to chunk audio for Avatars

PR #615, opened 2026-07-17
[Speechify] Add support for Speechify as a TTS provider

DaemonLoki:
## What's implemented?

- Speechify as a TTS provider, usable with `speechify.TTS()`
- We're supporting different models, up to their latest `simba-3.2`
- We have support for the different voice (e.g. `geffen_32` as the default, but others, like `beatrice_32`, etc)
- Multilingual support works as well, using e.g. `speechify.TTS(model="simba-multilingual", language="de-DE")`

## Demo

https://github.com/user-attachments/assets/b39b03d8-6944-4955-9dfd-5f8672ac5b9e

### DIFF
--- a/plugins/speechify/pyproject.toml
+++ b/plugins/speechify/pyproject.toml
@@ -1,0 +1,40 @@
+[build-system]
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[project]
+name = "vision-agents-plugins-speechify"
+dynamic = ["version"]
+description = "Speechify TTS integration for Vision Agents"
+readme = "README.md"
+keywords = ["speechify", "TTS", "text-to-speech", "AI", "voice agents", "agents"]
+requires-python = ">=3.10"
+license = "MIT"
+dependencies = [
+    "vision-agents",
+    "speechify-api>=1.2.4,<2",
+]
+
+[project.urls]
+Documentation = "https://visionagents.ai/"
+Website = "https://visionagents.ai/"
+Source = "https://github.com/GetStream/Vision-Agents"
+
+[tool.hatch.version]
+source = "vcs"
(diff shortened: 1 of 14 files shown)


plugins/speechify/pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+name = "vision-agents-plugins-speechify"
+dynamic = ["version"]
+description = "Speechify TTS integration for Vision Agents"
+readme = "README.md"
+keywords = ["speechify", "TTS", "text-to-speech", "AI", "voice agents", "agents"]
+requires-python = ">=3.10"
+license = "MIT"
+dependencies = [
+    "vision-agents",
+    "speechify-api>=1.2.4,<2",
```
luke-speechify:
```suggestion
    "speechify-api==3.0.1",
```

recommend jumping to 3.0.1 but i did break the wire when going to v3 (sorry!)

plugins/speechify/vision_agents/plugins/speechify/tts.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+
+        super().__init__(provider_name="speechify")
+
+        self.api_key = api_key or os.getenv("SPEECHIFY_API_KEY")
+        if not self.api_key:
+            raise ValueError("SPEECHIFY_API_KEY env var or api_key parameter required")
+
+        self.client = (
+            client if client is not None else AsyncSpeechify(api_key=self.api_key)
```
luke-speechify:
```suggestion
            client if client is not None else AsyncSpeechify(token=self.api_key)
```

i believe this is the breaking bit (again, sorry!)

luke-speechify (approved):

Nash0x7E2 (approved):

PR #617, opened 2026-07-27
fix(speechify): send Speechify-Caller attribution header

luke-speechify:
Adds headers={'Speechify-Caller': 'vision-agents'} to the AsyncSpeechify client construction in plugins/speechify/vision_agents/plugins/speechify/tts.py, plus a unit test asserting it. Matches the same attribution mechanic we've placed elsewhere.


DaemonLoki (approved):
Good catch, makes sense to add that!

dangusev (approved):

PR #636, opened 2026-08-19
fix: stabilize failing ElevenLabs, HuggingFace, and Sarvam tests

DaemonLoki:
## Why
Several plugin tests and model assumptions drifted from current provider behavior, which made CI flaky or fail outright. ElevenLabs STT can emit multiple utterances per clip, HuggingFace VLM interrupt coverage needed a more reliable signal than counting streamed deltas, and Sarvam no longer offers `sarvam-30b`.

## Changes
- Relax ElevenLabs STT turn-event assertions and add unit coverage for partial/committed transcript callbacks
- Make HuggingFace TransformersVLM interrupt tests wait on `model.generate` and assert truncated generation
- Remove unsupported `sarvam-30b` from `sarvam.LLM` and update README/tests to `sarvam-105b`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Breaking Changes
 
-### `sarvam` plugin: drop `sarvam-30b` from supported LLM models
+### `sarvam` plugin: drop `sarvam-30b` from supported LLM models (#636)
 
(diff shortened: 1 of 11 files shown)


Nash0x7E2:
@DaemonLoki can you check the conflicts here?

dangusev (approved):

PR #664, opened 2026-09-23
feat(sts): route OpenAI GPT-Live-1 over the Live API

tschellenbach:
## Why

GPT-Live-1 debuted at the top of the Artificial Analysis speech-to-speech index. It is not a Realtime model. OpenAI's model page says to "Use `gpt-live-1` in your API requests" and lists only the new Live endpoint, `wss://api.openai.com/v1/live/sessions`. The Live API has its own protocol: a `session.start` handshake and `session.*` events. It has no configurable turn detection and no image input. Instructions and voice cannot change after start. Reasoning and tools are delegated to a Responses backend model and run asynchronously while the voice keeps talking. The existing `openairealtime` package cannot drive it, so this needed code, not only a yaml entry.

**Do not merge yet: the live suite has not passed.** The `OPENAI_API_KEY` in this environment is rejected with 401 `invalid_api_key` by `/v1/live/sessions` and by `/v1/models`. Every test in the new `openailive` integration suite fails at the handshake with `bad handshake (http 401)`. The existing OpenAI Realtime suite fails the same way. The suite needs to be re-run with a valid key before this is trusted in production. Separately, the xAI suite that lives in the `openairealtime` package had one failure, "the model should have spoken". That code is untouched here.

GPT-Live-1 sits alongside gpt-realtime-2 and does not replace it. Live cannot see, has no semantic turn detector and no endpointing, and the `sts-vision` alias prefers `openai/gpt-realtime-2` and requires image input. No aliases or `prefer:` shortcuts were changed.

## Changes

- New `internal/sts/openailive` package, registered under the `openai` provider name. The registry picks it for `gpt-live-1*` models and keeps `openairealtime` for everything else. A Responses delegation carries the tools. The backend defaults to `gpt-5.6-terra`, the model OpenAI's delegation guide recommends, and can be changed through `backend_model` in overwrites.
- The Live API sends no reply-done event, so a reply is inferred from its audio and settled after a quiet gap. A reply is marked interrupted if the caller spoke over it. `Interrupt` mutes the rest of the reply locally, as the Gemini package does. `SendFrame` and `SetInstructions` refuse.
- `router.yaml` adds `openai/gpt-live-1` with capabilities `tools`, `text`, `input_transcript` and `output_transcript`; no `semantic_turns` and no `image`. The tier is `high-quality` and languages are `[en, pt]`, the only languages OpenAI names voices in.
  - Data policy: `trains_on_data: "no"` and `retention: unknown`. The Live API is missing from OpenAI's per-endpoint retention table, and they do not say what the backend Responses calls retain.
  - Price: `per_audio_hour: 3.0`, from $0.05 per minute billed per second. Backend tokens are billed on top and not counted, so this undercounts.
- Pricing: Artificial Analysis puts GPT-Live-1 at $4.47–5.83 per input-audio hour once backend tokens are included, above gpt-realtime-2 High at $4.14. So gpt-realtime-2's description changes from "The dearest by far." to "Among the dearest."
- Time to first audio: Artificial Analysis measures 1.24–1.34 s. That is slower than gpt-realtime-2, although OpenAI's announcement claims 0.8 s.
- Test harness: the stsrouter test stub now takes capabilities per model rather than per provider name, because one provider name now serves two packages. Added a test that the default registry builds a Live provider for `gpt-live-1` and a Realtime one for `gpt-realtime-2`.
- Pre-existing and out of scope: in workspace mode, `go build ./...` fails because `benchmark/go.mod` pins a `getstream-go-webrtc` revision that no longer exists. With `GOWORK=off`, `internal/agent/streamedge` does not compile (`videosdk.NewClientWithSecret` is undefined). Both failures happen identically on `accelerate`. The router-sts skill table does not yet describe the Live API.

<div><a href="https://cursor.com/agents/bc-4a6da6c1-7d1b-42df-895b-77300a11fcbf?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4a6da6c1-7d1b-42df-895b-77300a11fcbf&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


tschellenbach:
Ran the live suite locally against a valid `OPENAI_API_KEY`, since the cloud secret this branch was built with returns 401. The 401 is gone, but the suite does not pass, so this is not ready to merge yet.

`gpt-live-1`, `gpt-5.6-terra` and `gpt-6-astra` all exist on the account, so the model ids in this branch are right and the backend default is reachable. The failures are in the package, not the config.

Results of `go test -tags integration -run TestOpenAILiveIntegrationSuite ./internal/sts/openailive`:

| Test | Result |
| --- | --- |
| `TestTheModelHearsTheCallerAndAnswersInItsOwnVoice` | FAIL — the model speaks, but `reply.Complete.Interrupted` is true |
| `TestAPromptMakesTheModelSpeakUnasked` | FAIL — timed out after 120s |
| `TestATypedTurnIsAnswered` | FAIL — timed out after 120s |
| `TestAToolIsCalledAndItsAnswerCarriesTheConversationOn` | FAIL — timed out after 120s |
| `TestCuttingInStopsTheReplyAndSettlesItOnce` | PASS |
| `TestInstructionsChangeOnlyWhereTheModelAllowsIt` | PASS |

Two distinct problems, and they line up with the two design notes in the PR description.

The first is the settle heuristic. Audio does arrive, so the socket and the session setup are fine, but the reply settles as interrupted when nobody cut in. That points at "a reply counts as finished after a short quiet gap in its audio" firing during a gap inside the reply rather than at the end of it. Note that `TestCuttingInStopsTheReplyAndSettlesItOnce` passes, so a real barge-in is detected correctly — it is the un-interrupted case that is wrong.

The second is backend delegation. Every test that needs the Responses backend to produce the turn — an unprompted greeting, a typed turn, a tool round — waits the full 120s and hears nothing. The three that pass are the ones served by the live model alone. So the live leg works and the delegation leg does not.

The unit and fake-socket suites both pass, including under the race detector, which is consistent with this: the fake server asserts what we send, not what OpenAI actually answers.

tschellenbach:
## Fixed, and the whole live suite now passes

Followed up on the three timeouts I reported above. They were not three bugs — they were one, plus a wrong event.

**The Live session runs on a media clock.** Input audio is what advances it, and an append is injected at a point on that timeline. While the session is silent, nothing the application appends is ever reached, acknowledged or spoken. That is why every text-driven path hung and no path driven by real caller audio did.

Proved it against the API rather than inferring it. The same `session.commentary.append` that produces silence and no acknowledgment in a quiet session:

```
>> {"type":"session.commentary.append","delegation_id":null,"content":"Greet the caller warmly..."}
=== audio frames 0, said "" ===
```

is acknowledged and answered once 20ms silence frames are flowing:

```
<< {"type":"session.commentary.appended","start_ms":800,"end_ms":1000}
=== audio frames 149, said " Hi, thanks for calling—I'm here and happy to help." ===
```

With the clock running, the typed turn speaks (129 frames) and the tool round emits `function_call` with `{"city":"Paris"}`. **Every protocol shape in the original PR was correct** — including `delegation_id` being required on the appends and rejected on the `response.*` events, which I confirmed both ways.

### What changed

- The paths that append something and then wait to hear it carry the clock with 20ms silence frames until the model answers or the caller speaks. The filler stands down for both; an always-on filler spliced silence into the middle of the caller's utterance and broke the one test that had been passing.
- `Prompt` now sends `session.commentary.append` as well. An instruction only steers the speech a turn would have produced; the commentary is what starts the model talking. This is the pair OpenAI's greeting flow sends.
- A reply is no longer settled as interrupted because the caller was heard during it. This API transcribes the caller asynchronously, so the deltas for the utterance being answered normally land after the answer has begun — which made an ordinary reply report itself as cut off. Only `Interrupt` settles a reply as cut off now.

The unit test that asserted the old interrupted behaviour passed only because the fake server ordered its frames the way the real one does not. It now asserts the real behaviour.

### Verification

`internal/sts/openailive` unit tests pass 10/10 under `-race`. The live suite passes three consecutive runs in ~51s, against the previous 511s of timeouts:

```
--- PASS: TestAPromptMakesTheModelSpeakUnasked (5.96s)
--- PASS: TestAToolIsCalledAndItsAnswerCarriesTheConversationOn (6.10s)
--- PASS: TestATypedTurnIsAnswered (5.78s)
--- PASS: TestCuttingInStopsTheReplyAndSettlesItOnce (15.57s)
--- PASS: TestInstructionsChangeOnlyWhereTheModelAllowsIt (0.84s)
--- PASS: TestTheModelHearsTheCallerAndAnswersInItsOwnVoice (17.03s)
```

CI's remaining red is inherited from `accelerate` (`internal/session` and `internal/simulation`), not from this branch.

